PR #1, opened 2019-11-08
goinstant demo and organization changes based on OHIE2019 feedback

citizenrich:


citizenrich:
Found an error with Go. Will revisit and do another PR

PR #6, opened 2020-02-03
OHIE-310 config service

MattyJ007:


core/docker/importer.sh
```diff
@@ -1,2 +1,23 @@
...
-curl -X POST --insecure -u "$OPENHIM_API_USERNAME:$OPENHIM_API_PASSWORD" -H "Content-Type: application/json" -d @openhim-import.json "$OPENHIM_API_SERVER/metadata" -v
\ No newline at end of file
+OPENHIM_RESPONSE="";
+TIMEOUT=0
+
+while [ "$OPENHIM_RESPONSE" != "200" ]
+do
+  TIMEOUT=$((TIMEOUT+1))
+  # Fail script if incomplete at 30 seconds
+  if [ "$TIMEOUT" == "15" ]
```
BMartinos: The comment and the actual condition doesnt seem to match up. The comment states 30 seconds but the condition for the timeout is set at 15 seconds.

Assuming this has to do with the `sleep 2` which would equate to 30 seconds (15 iterations * 2 seconds sleeping). This however doesnt take into account the time it takes for the actual curl request to be completed.

Maybe this should be actual time based and not based on iterations

core/docker/importer.sh
```diff
@@ -1,2 +1,23 @@
...
+  fi
+
+  echo "OpenHIM not ready ( $OPENHIM_RESPONSE ) - sleeping"
+  sleep 2
+  OPENHIM_RESPONSE=$(curl -X GET --insecure --write-out %{http_code} --silent --output /dev/null "$OPENHIM_API_SERVER/heartbeat");
+done
+
+echo -e "\nSTART Posting OpenHIM Config\n----------------------------\n"
+curl --insecure -u "$OPENHIM_API_USERNAME:$OPENHIM_API_PASSWORD" -H "Content-Type: application/json" -d @openhim-import.json "$OPENHIM_API_SERVER/metadata" -v
+echo -e "\n\n--------------------------\nEND Posting OpenHIM Config\n"
```
BMartinos: Maybe we need to interpret the response here to inform the user of the status. If the json payload isnt formatted correctly, the final response will still print out that the OpenHIM config has been imported

core/docker/importer.sh
```diff
@@ -1,2 +1,23 @@
...
+  if [ "$TIMEOUT" == "15" ]
+  then
+    echo -e "Sleep timeout reached: $TIMEOUT.\nFailed to import config. Could not reach the OpenHIM."
+    exit 1
+  fi
+
+  echo "OpenHIM not ready ( $OPENHIM_RESPONSE ) - sleeping"
+  sleep 2
+  OPENHIM_RESPONSE=$(curl -X GET --insecure --write-out %{http_code} --silent --output /dev/null "$OPENHIM_API_SERVER/heartbeat");
+done
```
rcrichton: This looks like it would work well, however, we might use this same script to configure a number of services. I wonder if we should include a utility like this - https://www.npmjs.com/package/wait-on - in our `Dockerfile.importer` and rather use that to wait-on all services to be available. Have a look at the tool about and let me know your thoughts.

core/docker/importer.sh
rcrichton: I'm wondering if we should use a generic utility for this so we don't have to worry too much about this logic. See my comment below.

core/docker/importer.sh
rcrichton: I'm sure similar tools exist if you want to try find a bash only one, but node doesn't add much weight to the image.

BMartinos: Thanks @rcrichton Thats a good idea. We could make use of the `node-alpine` base docker image to keep things small. Maybe we should be changing this importer script to rather be a node application which will handle the importing logic? This could also easiliy be updated to cater for other import methods

What do you think? Reason for initially using bash is its quick and easy to get something basic working. More complex flows are def possible but can become quite confusing to follow and maintain.

BMartinos:
@rcrichton For these services, do you think we need to lock down services that shouldn't be accessible directly? Im thinking in terms of FHIR, shouldnt be accessible directly and should rather come from the IOL. As well as things like database, should they only be accessible from within the network?

core/docker/importer.sh
rcrichton: Yeah, transitioning this to a js script might be good. I don't have a problem between bash or js really. Either is a good option. We could just use the wait-on CLI in a bash script.

js might be easier for us to maintain so that is a plus.

rcrichton:
@BMartinos yeah I think they should be locked down. We could provide extension docker-compose files to expose all ports for development purposes like I describe in this unreleased documentation - https://jembi.github.io/architecture/docs/deployment#managing-docker-compose-files-for-multiple-environments

Then you can run a dev env with port exposed like this: `docker-compose -f docker-compose.yml -f docker-compose.dev.yml`

BMartinos:
@rcrichton great thanks. We will create a ticket for locking down these services as then to allow them to be overwritten for development purposes

core/docker/importer/openhim.js
```diff
@@ -0,0 +1,85 @@
...
+    resources: [`https://${OPENHIM_API_HOSTNAME}:${OPENHIM_API_PORT}/heartbeat`],
+    delay: 1000, // initial delay in ms, default 0
+    interval: 2000, // poll interval in ms, default 250ms
+    timeout: 30000, // timeout in ms, default Infinity
+    tcpTimeout: 1000, // tcp timeout in ms, default 300ms
+    window: 1000, // stabilization time in ms, default 750ms
+    auth: {
+      user: OPENHIM_API_USERNAME,
+      pass: OPENHIM_API_PASSWORD
+    },
```
BMartinos: do we need to supply the auth here for checking the heartbeat? The API doesnt require any auth on this endpoint

MattyJ007: No auth needed.
Fixed in https://github.com/openhie/instant/pull/6/commits/16c1169f6e260ccb42dab30f2ac0c1a7533ed998

PR #9, opened 2020-02-10
add scafold for the documentation website

BMartinos:
OHIE-315


docs/.docusaurus/client-modules.js
```diff
@@ -0,0 +1,4 @@
+export default [
+  require("infima/dist/css/default/default.css"),
+  require("/home/bradford/Developer/instant/docs/src/css/custom.css"),
```
BMartinos: This wont work on other developers computers. I also think we can ignore the `.docusaurus` directory. Lets create a gitignore to exclude these from the repo
https://github.com/jembi/openhim-docs/blob/master/.gitignore

bradsawadye: fixed https://github.com/openhie/instant/pull/9/commits/5a61fd1160b5db992bd221d8a65b58f4ac1eacc6

docs/sidebars.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = {
+  'docs': {
+    'Introduction': [
+      'introduction/overview',
+      'introduction/community'
+    ],
+    'Architecture': [
+      'architecture/overview'
+    ],
+    'Components': [
```
rcrichton: I think we should rename 'Components' to 'Packages' like it is in the architecture doc. Each package will contain one or more components.

BMartinos: Great thanks Ryan. Think this was some confusion from my side

bradsawadye: fixed in https://github.com/openhie/instant/pull/9/commits/a4d9f6fc9d902d1c89579cdeacd8d70ed79e1347

PR #10, opened 2020-02-11
OHIE-319 - add architecture docs and scaffold additional doc sections

rcrichton:


docs/sidebars.js
```diff
@@ -9,12 +9,15 @@ module.exports = {
...
       'introduction/community'
     ],
-    'Architecture': [
-      'architecture/overview'
+    'Concepts': [
+      'concepts/overview',
+      'concepts/package-arch'
     ],
     'Packages': [
+      'packages/creating-packages',
```
BMartinos: I wonder if this shouldn't rather be within the concepts section? this relates to how to create more packages which could be seen as a concept which can be approached in a few ways. The packages section contains all the currently built packages for the community to make use of

docs/src/pages/index.js
```diff
@@ -38,17 +38,17 @@ const features = [
...
-        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-        mollit anim id est laborum
+        Instant OpenHIE provides an easy way to setup, explore and develop with
+        the OpenHIE Architecture. It allows packages to be added that support
+        multiple different use cases and workflows specified by OpenHIE. Each
+        package contains scripts to stand up and configure applications that
+        support these various workflows.
+      <br /><br />
+        View the <a href="docs/concepts/overview">Concepts section</a> to learn more.
```
BMartinos: Maybe we can move this box into the middle to keep consistency with the sidebar

docs/sidebars.js
rcrichton: Yeah I was wondering about that to, was also considering a How-tos section, but didn't think we had enough to go in there. Cool, let me move it to concepts and see how it feels.

docs/src/pages/index.js
rcrichton: Good idea.

docs/docs/concepts/package-arch.md
```diff
@@ -0,0 +1,24 @@
...
+
+Each package will contain the following sorts of technical artefacts:
+* Docker compose scripts for setting up the applications required for this package’s use cases and workflows
+* Kubernetes scripts for setting up the applications required for this package’s use cases and workflows
+* Configuration scripts to setup required configuration metadata
+* Extensions to the test harness to test the added use cases with test data
+
+The below diagram shows how packages will extend off each other to add use cases of increasing complexity.
+
+![](package-arch.png)
```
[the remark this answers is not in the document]
rcrichton: Cool, I've made the change. The one issue I have with doing it this way is that it now doesn't render well if you navigate to the markdown page on Github and in vscode when previewing where as it will if done using normal markdown format. It's not a huge deal though, seems it's the recommended approach.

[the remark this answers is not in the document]
BMartinos: Yeah it doesnt make the preview of these docs look very nice as it doesnt render the images. I also dont like that part as its harder to see the layout. But maybe this should ideally be viewed in the context of the website and not individually.

The main reason for using the context is to ensure the path always points to the images regardless of where its deployed. Currently we have the `/instant/` path set as its being served under openhie github, but if we had a sub-domain instead the baseUrl would just be `/` and any hardcoded references would be broken

PR #12, opened 2020-02-12
OHIE-318: Initial introduction docs

BMartinos:
So that when we do a deploy it goes through sucesssfully

OHIE-318


docs/docusaurus.config.js
```diff
@@ -1,12 +1,17 @@
...
+  url: 'https://openhie.github.io',
+  baseUrl: '/instant/',
   favicon: 'img/favicon.ico',
-  organizationName: 'OpenHIE',
-  projectName: 'Instant OpenHIE',
+  organizationName: 'openhie',
+  projectName: 'instant',
   themeConfig: {
+    algolia: {
+      apiKey: '',
```
BMartinos: The key has been requested and awaiting a response from Algolia

PR #18, opened 2020-02-18
OHIE-332 - k8s core package config job

rcrichton:


core/docker/docker-compose.config.yml
```diff
@@ -3,10 +3,12 @@ version: '3.3'
 services:
   # container for executing config import scripts for setting up initial config
   config-importer:
-    container_name: openhim-config-importer
+    container_name: instantohie-config-importer
```
BMartinos: We might need to revert this back to be specific to the package creating this container.
If we have another package also making use of this image for implementing config we wont be able to use the same container name. This way we also keep it specific to the package its performed on

core/kubernetes/config/config-importer-job.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: config-importer
+spec:
+  template:
+    spec:
+      containers:
+      - args:
+        - sh
+        - -c
+        - wait-on -t 30000 https-get://openhim-core:8080/heartbeat && node openhimConfig.js
```
BMartinos: Just a note: This might need to be changed later on when we have the consolidated service names that we will be using. This might even also be a custom domain name setup with Ingress

core/kubernetes/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+configMapGenerator:
+  - name: importer-config
+    files:
+    - config/volume/openhim-import.json
+    - config/volume/openhimConfig.js
```
BMartinos: minor: Think this should be indented

PR #22, opened 2020-02-21
OHIE-329 kubernetes deploy scripts

MattyJ007:


core/kubernetes/importer/core-openhim-config-importer-job.yaml
```diff
@@ -12,6 +12,10 @@ spec:
   backoffLimit: 0
   template:
     spec:
+      initContainers:
+        - name: check-openhim-core-status
+          image: busybox:1.28
+          command: ['sh', '-c', 'until telnet openhim-core-service 8082; do echo OpenHIM Core API not running yet; sleep 10; done;']
```
rcrichton: The importer already implementing it's own waiting. Should we scrap that for this instead?

BMartinos: @rcrichton hmm good question. I think i added this in because i ran into an issue when on first startup the config it took longer than 5 minutes to start and had the container in a crashed state. Maybe we dont need both. Ill do a clean run install again to see if this is needed. Which method do you prefer for waiting on services?

I kind of like the initContainer approach as the container logic isnt started and restarted when it crashes

rcrichton: I kinda like Init Containers to. That said we still need a way for this config to work for the docker compose setup. I don't think there is anything equivalent in docker. Maybe that can be a reason to keep both? I'm not sure. What do you think?

BMartinos: Yeah i would maybe say lets make use of the initContainer where possible to keep consistency throughout. For docker we will need to make use of this waitOn scrpt as there isnt a different way to achieve the same result

PR #23, opened 2020-02-24
Ohie 331 dev scripts

bradsawadye:


core/kubernetes/README.md
```diff
@@ -29,3 +29,14 @@ Once you are in the correct working directory (`core/kubernetes/`) we can procee
 ./setup.sh
 ```
 
+To run in development mode, where the OpenHIM mongo database, HAPI fhir server and the MySQL database can be accessed directly through their urls, run the following command
+
+```bash
+./setup.sh dev
```
BMartinos: I dont think we should have the `dev` flag as part of the `main/setup.sh` script. This script should be in its own `dev` folder. We could add a question in this bash script to ask if they have started the `main` scripts so that this can be executed on top of the existing services

core/kubernetes/dev/dev-urls.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+res="$(kubectl get service openhim-mongo-dev-service)"
+
+if [ "$?" == "0" ]; then
+  url="$(minikube service openhim-mongo-dev-service --url)"
```
BMartinos: We might need to replace the protocol (`http://`) from this result as the database is accessed via tcp and not http

core/kubernetes/dev/dev-urls.sh
```diff
@@ -0,0 +1,23 @@
...
+
+  echo -e "The mongo url:\n $url\n"
+else
+  echo "Mongo dev not started (in dev mode)"
+fi
+
+res1="$(kubectl get service hapi-fhir-mysql-dev-service)"
```
BMartinos: We might need to replace the protocol (`http://`) from this result as the database is accessed via tcp and not http

core/kubernetes/dev/dev-urls.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+res="$(kubectl get service openhim-mongo-dev-service)"
```
BMartinos: We should name this variable a bit more descriptive in terms of what response it is.

core/kubernetes/dev/dev-urls.sh
BMartinos: We should name this variable a bit more descriptive in terms of what response it is.

core/kubernetes/main/k8s.sh
```diff
@@ -3,6 +3,14 @@
...
     kubectl apply -f ./openhim
     kubectl apply -f ./hapi-fhir
+
+    if [ "$2" == "dev" ]; then
+        kubectl apply -f ../dev/openhim
+        kubectl apply -f ../dev/hapi-fhir
+        
+        echo -e "\nCurrently in development mode\n"
+    fi
+
```
BMartinos: If we create the bash script for the dev setup then this can be removed as this will be included in the `setup.dev.sh` script

core/kubernetes/main/setup.sh
```diff
@@ -4,5 +4,5 @@ echo "Setup Kubernetes..."
 kubectl create configmap console-config --from-file=../config/openhim-console/default.json
 kubectl create configmap core-config --from-file=../config/openhim-core/development.json
 kubectl create configmap hapi-fhir-config --from-file=../config/hapi-fhir/hapi.properties
-./k8s.sh up
+./k8s.sh up "$1"
```
BMartinos: This can be reverted with the `dev` specific script

PR #27, opened 2020-02-26
OHIE-342 Add base docker and k8s config for mapping mediator

rcrichton:
This PR can be tested by executing:

For a docker deploy:
```bash
./instant.sh up
# wait until all the services are started, monitor using:
watch docker ps
# when all up:
./instant.sh test
```

For a kubernetes deploy:
```bash
./instant.sh up k8s
# wait until all the services are started, monitor using:
watch kubectl get all
# when all up:
./instant.sh test k8s
```


core/docker/compose.sh
```diff
@@ -0,0 +1,11 @@
+composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+
+if [ "$1" == "up" ]; then
```
bradsawadye: @rcrichton, are we always going to be in dev mode?

core/kubernetes/importer/core-openhim-config-importer-job.yaml
```diff
@@ -37,7 +37,7 @@ spec:
           volumeMounts:
             - mountPath: /importer
               name: core-openhim-config-importer-config
-      restartPolicy: Never
+      restartPolicy: OnFailure
```
MattyJ007: @rcrichton Why is this restart on OnFailure? Surely if it fails it means there's something wrong with the OpenHIM instance or we've supplied the wrong credentials. Either way, we would have triggered this manually and seen it fail. We would then need to redeploy the OpenHIM or the importer and in that time there would be no need to reattempt the importer job

core/docker/importer/docker-compose.config.yml
```diff
@@ -2,10 +2,10 @@ version: '3.3'
...
   # container for executing config import scripts for setting up initial config
-  config-importer:
+  core-config-importer:
     container_name: core-config-importer
     build:
-      context: importer
+      context: .
       dockerfile: Dockerfile
     image: jembi/instantohie-config-importer
     restart: "no"
```
bradsawadye: For consistency, I think we should restart the container on failure as we do for in k8s https://github.com/openhie/instant/pull/27/files/feb9e7c6888229af6e1f9b2144984d6a97a545f8#diff-db7d6f0d86d65ba0727595cb8e75692eR40

healthworkforce/docker/importer/volume/README.md
```diff
@@ -0,0 +1,3 @@
+# Note
+
+These are replicated from healthworkforce/docker/volume as Kubernetes need these to be under the path where it config files are executed from.
```
bradsawadye: typo `where it config`

healthworkforce/kubernetes/importer/volume/README.md
```diff
@@ -0,0 +1,3 @@
+# Note
+
+These are replicated from healthworkforce/docker/volume as Kubernetes need these to be under the path where it config files are executed from.
```
bradsawadye: typo `where it config`

bradsawadye: fixed in https://github.com/openhie/instant/pull/27/commits/2811a62e7f0ab413e1b4b50b9670e17a83400f70

healthworkforce/docker/importer/volume/README.md
bradsawadye: fixed in https://github.com/openhie/instant/pull/27/commits/2811a62e7f0ab413e1b4b50b9670e17a83400f70

core/kubernetes/importer/core-openhim-config-importer-job.yaml
rcrichton: We want all the setup to be as automatic as possible. We don't want to someone to be watching the status of the job. I think it's better to retry on failure. I've had cases where this failed because something wasn't up yet and the retry helps in this case.

core/docker/compose.sh
rcrichton: I'm thinking for the docker deploy we can assume dev mode for the moment. This is generally what it will be used for. I don't want to make the scripts too complex by supporting multiple modes as the might be replaced by the go app in the future.

core/kubernetes/importer/core-openhim-config-importer-job.yaml
rcrichton: Also, the ./instant.sh script executed both the normal and config script for you so the user might not be aware of needing to restart the job.

rcrichton: Sorry, another point :) The end goal of this project is for a lay person to be able to spin up an HIE automatically by the click of a button. So, as much as we can automate and make failure resilient the better.

BMartinos: Maybe just to jump in a bit. If the job fails for some reason it is most likely due to a valid reason in terms of not authenticating to the service its trying to import to.

The initContainer is used to verify that the service we are trying to reach is up and running and able to receive the request we are sending. This means the pod will not fail as it wont have started yet and will keep trying. 

If the pod fails at this point then its most likely the user credentials which is incorrect, or the import object isnt valid. Restarting the same job at this point will not have a different result as the creds or object will still be invalid

BMartinos: What we maybe need is some kind of feedback to the user to inform them that the import failed.

Running the setup from scratch should not result in any errors as everything is setup vanilla. but if we trigger the entire setup script again it will fail on the importer due to the password having changed. If this job is set to restart it will never complete and therefor never complete the `instant.sh` startup process

rcrichton: I ran into cases where restarting on failure was necessary, but I think this was before the init container got merged. I don't mind trying to change these to `restartPolicy: Never` if you guys think that is best and then we can see if we encounter scenarios where this is necessary again.

I'l make the change for the moment.

rcrichton: I'm leaving the restart policy on the docker ones as we don't have a way to ensure those are up before these are run. Does that make sense to you?

rcrichton: Ok, I tested this out and the problem that I run into is that the mapper-config job can sometimes try to run before the core config has been run. This causes it to fail because the root user password is not yet set. We don't have a good way to make sure that job has run yet. The `restartPolicy: OnFailure` was a good stop gap for this. Do you think it should stay for this reason or should we try find another approach?

BMartinos: hmmm ok cool, i see how your scenario plays out. So the core API starts up which means `/heartbeat` is accessible but at this time the `root@openhim.org` hasnt been inserted into the DB yet, So the request to import fails as this user doesnt yet exist. 

In that case i think we need to include a backoff period, so that the pod is only restarted say maximum 3 times so that it cant get stuck in a failure state. 

We might also need to do a cleanup of the job once its completed to remove any pods it created that might be in an Error state.

https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#pod-backoff-failure-policy

BMartinos: Sorry realized that version of Kubernetes isnt yet supported on AWS.
https://v1-14.docs.kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

rcrichton: Alright, I've set the backoffLimit to 3. I didn't want to run a clean up automatically as then we lose logs that will be useful for any debugging.

BMartinos: Ok cool sounds good :+1:

PR #28, opened 2020-02-28
OHM-972 AWS

MattyJ007:

### DIFF
--- a/core/docker/compose.sh
+++ b/core/docker/compose.sh
@@ -1,7 +1,7 @@
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 if [ "$1" == "up" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml  up -d
+    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml up -d
 elif [ "$1" == "down" ]; then
     docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml stop
 elif [ "$1" == "destroy" ]; then
--- a/core/docker/compose.sh
+++ b/core/docker/compose.sh
@@ -5,7 +5,7 @@ if [ "$1" == "up" ]; then
 elif [ "$1" == "down" ]; then
     docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml stop
 elif [ "$1" == "destroy" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml down
+    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml down -v
 else
     echo "Valid options are: up, down, or destroy"
 fi
--- a/core/kubernetes/README.md
+++ b/core/kubernetes/README.md
@@ -9,6 +9,13 @@ This package consists of two services:
 
 ## Kubernetes Host Environment Setup
 
+> If you are working with multiple cluster hosting clients you may need to switch cluster contexts for you deployments. See the below commands:
+
+```sh
+kubectl config get-contexts
+kubectl config use-context <context-name>
+```
+
 ### Minikube (local)
 
 For the Kubernetes deployment to work as expected, we need to ensure we have `minikube` installed on our local machine running the deployment. Follow these steps to [install minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/)
@@ -21,7 +28,12 @@ minikube start --cpus 4 --memory 8192
 
 This also updates the VM settings to make use of 4 CPU's and 8GB of RAM, instead of the default 2 CPU's and 4GB of RAM
 
-### AWS CLI
+### Amazon Web Services
+
+Useful links:
+
+- [AWS Cli Setup guide](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html)
+- [Amazon Elastic Kubernetes Service(EKS) Quickstart](https://docs.aws.amazon.com/eks/latest/userguide/getting-started-console.html)
 
 Some prerequisites are required before we can continue to deploy our Kubernetes infrastructure to an AWS cluster.
 
@@ -29,10 +41,6 @@ Some prerequisites are required before we can continue to deploy our Kubernetes 
 - You have given the users the relevant access to the AWS services
 - You have generated an access token for your AWS user
 - You have installed all the relevant CLI tools
-
-Useful links:
-
-- [EKS](https://docs.aws.amazon.com/eks/latest/userguide/getting-started-console.html)
 
 #### Install AWS Cli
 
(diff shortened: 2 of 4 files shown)


core/kubernetes/main/k8s-mini.sh
```diff
@@ -16,9 +19,6 @@ elif [ "$1" == "down" ]; then
     kubectl delete deployment hapi-fhir-server-deployment
     kubectl delete deployment hapi-fhir-mysql-deployment
 elif [ "$1" == "destroy" ]; then
-    # delete host entry on destroy
```
BMartinos: Was there a reason for removing this in the destroy? This just allows for cleanup of the host file so that the user doesnt have this entry in there forever

core/kubernetes/main/k8s-aws.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ "$1" == "up" ]; then
+    kubectl apply -k $kustomizationFilePath
+
+    openhimApiUrl=$(kubectl get service openhim-core-service -o=jsonpath={.status.loadBalancer.ingress[0].hostname})
+    fhirServerUrl=$(kubectl get service hapi-fhir-server-service -o=jsonpath={.status.loadBalancer.ingress[0].hostname})
+
+    printf "\n\nOpenHIM Api Url\n---------------\n"$openhimApiUrl"\n"
+    printf "\nHAPI FHIR Url\n--------------\n"$fhirServerUrl"\n\n\n"
+
+    sed -i -E "s/(\"host\": \")\S*(\")/\1${openhimApiUrl}\2/" $openhimConsoleVolumePath
```
BMartinos: Maybe add a comment here to say that this is injecting the relevant service URL into the config

core/kubernetes/main/k8s-mini.sh
```diff
@@ -6,6 +6,9 @@ if [ "$1" == "up" ]; then
     minikube addons enable ingress
     kubectl apply -k $kustomizationFilePath
 
+    kubectl get services
+    kubectl get ingress
```
rcrichton: Are these just for display purposes?

core/kubernetes/main/openhim/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+    files:
+      - ./volume/openhim-console/default.json
+
+resources:
+  # deployment
+  - ./openhim-console-deployment.yaml
+  # service
+  - ./openhim-console-service.yaml
+  # ingress
+  - ./openhim-console-ingress.yaml
```
rcrichton: I think we should have a single kustomization file for the core package. The idea with packages is that they can be stood up and torn down individually. So, I'm not sure why we would want to split out the OpenHIM to be setup on it's own.

core/kubernetes/main/k8s-aws.sh
```diff
@@ -0,0 +1,37 @@
...
+openhimConsoleVolumePath="${kustomizationFilePath}/openhim/volume/openhim-console/default.json"
+openhimApiUrl=''
+fhirServerUrl=''
+openhimConsoleUrl=''
+
+if [ "$1" == "up" ]; then
+    kubectl apply -k $kustomizationFilePath
+
+    openhimApiUrl=$(kubectl get service openhim-core-service -o=jsonpath={.status.loadBalancer.ingress[0].hostname})
+    fhirServerUrl=$(kubectl get service hapi-fhir-server-service -o=jsonpath={.status.loadBalancer.ingress[0].hostname})
```
rcrichton: Surely similar techniques would work with minikube to? Could we not rather try update the original script to work with eks rather than having two very similar scripts. That will make it easier for us to maintain and also won't break the `./isntant.sh` script which will be the entry point for running all of this.

core/kubernetes/main/openhim/kustomization.yaml
rcrichton: Hmm, on second though is this done this way because we need to get the urls before we can configure the console?

rcrichton:
Another general note, we want to try avoid anything that directly ties us to AWS EKS specifically. We may want to deploy to a different managed Kubernetes environment. It's cool to have docs that help us get EKS setup, but our script should try to be environment agnostic.

core/kubernetes/main/k8s-mini.sh
MattyJ007: This was an accident :|

core/kubernetes/main/openhim/kustomization.yaml
MattyJ007: Yeah exactly. Can't have the console start before the changes have been made to it config files

core/kubernetes/main/k8s-mini.sh
MattyJ007: Yeah - but they're actually not that useful at this point

BMartinos:
@rcrichton I do agree with you comment about getting this to a point where its environment agnostic but think maybe initially that will over complicate things. 

Maybe we should first get focused on getting this all working as intended on a specific stack and then in future look at other stacks? It will be a bit hard to verify this would work on other stacks without getting access to those environments. Setting up Azure / Google Cloud projects and upskilling on how to deploy to these

Even if this is agnostic of environment, we still need to highlight what is needed to deploy to those stacks. Maybe this is something we can look at later on as this is more mature? What do you think?

rcrichton:
@BMartinos I agree in principle, but there are some thing we can do to try be as environment agnostic as we can and only add it extra stuff where absolutely necessary. If there are parts of the scripts that will be the same we should look for way to re-use and modify instead of create entirely new ones, this will make our job easier down the line. I agree with being pragmatic but lets bring them together as much as we can without making things too complex.

Relating to the above, I think the k8s.sh script is mostly the same and can be re-used with only a few tweaks.

BMartinos:
Thanks @rcrichton thinking about this some more. I think it does make sense to have all this logic sit in one `k8s.sh` script. 

There are however some differences in deploying to cloud/minikube which we will need to change to keep them as closely aligned as possible.

We are thinking of removing the ingress related files as this only benefited the minikube installation to not have to worry about updating services with other services URLs. This has since been fixed in the deploy script to extract and update these where relevant to ensure all transactions flow successfully.

The `k8s.sh` script will first determine the environment based on the current `kubectl config get-contexts` to which environment to deploy to and configuring anything specific to that environment

rcrichton:
@BMartinos ya, that sounds better and I think it's the way to go for now.

**Sidenote:** I'm not convinced that getting the kubenetes urls is the best way to get the public urls. We may want to use ingress on AWS to but that requires us to have a domain name. This would make it possible for us to know the URL up front and deploy everything in one go again (rather than deploying the console separately). I'm just not sure how this would work in AWS as the IP addresses that the DNS record points to might change each time we create a cluster. It's something we can figure out in the future perhaps. Just letting you know my thinking.

BMartinos:
@rcrichton correct, Having ingress will be great to have the domain names predefined which will speed up the deploy process as we wont have to wait on services. Unfortunately ingress wont work in the cloud until we have a domain name associated with the cluster. This will however not work for other people trying to deploy these scripts as they might not have a domain available.

If we make use of the service IP then these shouldnt change for the cluster, unless the cluster (or services) are destroyed and re-created. Its definitely something to keep in mind for future if we ever need to take these scripts to a production ready state.

Making use of the IPs isnt ideal but its the most consistent way to deploy to both minikube and the cloud without significantly different steps

instant.sh
```diff
@@ -45,8 +45,19 @@ if [ "$TARGET" == "kubernetes" ] || [ "$TARGET" == "k8s" ]; then
         ./healthworkforce/kubernetes/main/k8s.sh destroy
         ./healthworkforce/kubernetes/importer/k8s.sh clean
     elif [ "$COMMAND" == "test" ]; then
-        ./core/test.sh $minikubeIP:$openhimCoreTransactionSSLPort
-        ./healthworkforce/test.sh $minikubeIP:$openhimCoreTransactionSSLPort
+        openhimCoreHostname=$(kubectl get service openhim-core-service -o=jsonpath="{.status.loadBalancer.ingress[*]['hostname', 'ip']}")
+        openhimCoreTransactionSSLPort=$(kubectl get service openhim-core-service -o=jsonpath={.spec.ports[1].port})
+        hostnameLength=$(expr length "$openhimCoreHostname")
+        echo "Host: $openhimCoreHostname"
+        echo "Length: $hostnameLength"
```
rcrichton: Is this maybe testing output or do you want this here?

MattyJ007: Whoops. Testing :(

core/docker/compose.sh
```diff
@@ -1,7 +1,7 @@
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 if [ "$1" == "up" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml  up -d
+    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml up -d
 elif [ "$1" == "down" ]; then
     docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml stop
 elif [ "$1" == "destroy" ]; then
```
MattyJ007: @rcrichton Is there an easy way of removing the volumes associated with these containers on destroy?

I tried using the method described here:
https://docs.docker.com/compose/reference/rm/

But it didn't seem to work? Every time I run `docker-compose up` it created a new anonymous volume that is never cleaned up

MattyJ007: It might be easier to clean these up if we named the volumes and deleted them specifically...

rcrichton: Ya sure can do that. I also see the `docker-compose rm` docs state: By default, anonymous volumes attached to containers are not removed. You can override this with -v

MattyJ007: I tried that. Didn't appear to work

rcrichton: Ok, lets switch to named volume then, it's probably better. Are you ok to do this else I can look later today?

MattyJ007: I'll get on it :+1:

MattyJ007: Tried it again from a completely fresh start and it worked :| 
Think I must have had some other existing stuff mess with it :( 
But solved :D

core/kubernetes/README.md
```diff
@@ -9,6 +9,13 @@ This package consists of two services:
 
 ## Kubernetes Host Environment Setup
 
+> If you are working with multiple cluster hosting clients you may need to switch cluster contexts for you deployments. See the below commands:
```
bradsawadye: typo: contexts for you deployments

PR #34, opened 2020-05-04
Docker outside of Docker and K8s outside of Docker exploration

rcrichton:
This PR show how the `instant.sh` script and each packages `compose.sh` and `k8s.sh` script can be used cross platform via a docker orchestration container.

This PR is for demonstration and discussion purposes only, it need more work to ensure full compatibility and to properly function under all conditions.


rcrichton:
See https://github.com/openhie/instant/pull/48

PR #37, opened 2020-05-07
OHIE-379 Change the link color to the brand's blue

MattyJ007:
The brand green was difficult to read on the white background

OHIE-379

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -6,13 +6,13 @@
 
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-color-primary: #00d097;
+  --ifm-color-primary: #0e67d8;
   --ifm-color-primary-dark: #0e67d8;
   --ifm-color-primary-darker: #083ea7;
   --ifm-color-primary-darkest: #302f3d;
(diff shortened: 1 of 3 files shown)


docs/src/css/custom.css
```diff
@@ -35,3 +35,8 @@
   margin: 0 calc(-1 * var(--ifm-pre-padding));
   padding: 0 var(--ifm-pre-padding);
 }
+
+.button.button--green {
```
rcrichton, line 39: `button--green` might be a bit of a confusing name. Maybe `button--primary` would be better?

MattyJ007: That would override the primary style that exists. Happy to change it to another name though.
I used the descriptor green as buttons will be default be blue due to docusaurus default styling. I don't want to override the primary stylings at all because the green won't work well on a white background. (Not that we currently have other buttons...)

MattyJ007: @rcrichton ?

rcrichton: Cool, anything that doesn't mention the colour would be good. Just think we should decouple the class from the colour. Not a big deal though.

PR #41, opened 2020-05-12
Ohie 367 mock services for testing

bradsawadye:


healthworkforce/mock-services/mcsdMockServices/resources/bundle-transaction-response.json
```diff
@@ -0,0 +1,126 @@
...
+        "status": "200 OK"
+      }
+    },
+    {
+      "response": {
+        "status": "304 Not Modified"
+      }
+    }
+  ]
+}
```
MattyJ007, line 126: Add new line

healthworkforce/mock-services/mcsdMockServices/src/logger.js
```diff
@@ -0,0 +1,19 @@
+const pino = require('pino')
```
MattyJ007, line 1:
```suggestion
'use strict'

const pino = require('pino')
```

healthworkforce/mock-services/mcsdMockServices/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+  logger.info(`Fetching iHRIS practitioner data with query paramter`)
+  res.json(practitionerResource)
+})
+
+app.get('/ihris-practitionerRole-mock/_history', function(req, res) {
+  logger.info(`Fetching iHRIS practitionerRole data with query paramter`)
+  res.json(practitionerRoleResource)
+})
+
+app.post('/fhir-mock', function(req, res) {
```
MattyJ007, line 52: In JS if a proceeding parameter is not used you can prefix it with an underscore to indicate this to other developers and to the linter.
```suggestion
app.get('/gofr-location-mock/_history', function(_req, res) {
  logger.info(`Fetching GOFR Location data with query paramter`)
  res.json(locationResource)
})

app.get('/gofr-organization-mock/_history', function(_req, res) {
  logger.info(`Fetching GOFR Organization data with query paramter`)
  res.json(organizationResource)
})

app.get('/ihris-practitioner-mock/_history', function(_req, res) {
  logger.info(`Fetching iHRIS practitioner data with query paramter`)
  res.json(practitionerResource)
})

app.get('/ihris-practitionerRole-mock/_history', function(_req, res) {
  logger.info(`Fetching iHRIS practitionerRole data with query paramter`)
  res.json(practitionerRoleResource)
})

app.post('/fhir-mock', function(_req, res) {
```

core/docker/compose.sh
```diff
@@ -1,6 +1,11 @@
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 if [ "$1" == "up" ]; then
+    docker-compose -f "$composeFilePath"/docker-compose-mongo.yml up -d
+
+    # Set up the replica set
+    "$composeFilePath"/initiateReplicaSet.sh
```
MattyJ007, line 7: @BMartinos @rcrichton could we add another startup option? `init`. This option could contain the Importer script and the Mongo init scripts. 

Currently, if we were to stop the Instant services and then try to use `up` to start them again there will be problems - this is an existing issue as the importer script would have silently failed on the re-use of `up`. But now we are going to re-initialize the Mongo replica set. This shouldn't fail - but it is unnecessary.

BMartinos: @MattyJ007 how would the `init` option work in practice? Normal compose first and then init? There might be some dependencies making this a bit harder to resolve. For replica set to work we need to mongos running, For the importer to work we need the core stored, which needs mongo to work properly

MattyJ007: Init would basically replace `up` and up would then just be `docker-compose up -f ...` (no importer script or replica specific stuff)

BMartinos: Ok cool i see. So its a duplicate of the `up` option which contains all the scaffolding related stuff. And then the `up` option will just do the startup and no configuration setup. I think that sounds fine to me :+1:

core/docker/compose.sh
```diff
@@ -1,6 +1,11 @@
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 if [ "$1" == "up" ]; then
+    docker-compose -f "$composeFilePath"/docker-compose-mongo.yml up -d
```
MattyJ007, line 4: @bradsawadye This `docker-compose-mongo.yml` needs to be referenced in the `down` and `destroy` sections of the script else it won't be cleaned up properly

core/docker/compose.sh
bradsawadye: addressed in https://github.com/openhie/instant/pull/41/commits/26046074e16c00c733162602fec1a77fc888941b

healthworkforce/mock-services/mcsdMockServices/resources/bundle-transaction-response.json
bradsawadye: addressed in https://github.com/openhie/instant/pull/41/commits/26046074e16c00c733162602fec1a77fc888941b

core/docker/compose.sh
bradsawadye: addressed in https://github.com/openhie/instant/pull/41/commits/26046074e16c00c733162602fec1a77fc888941b

core/docker/compose.sh
```diff
@@ -1,11 +1,23 @@
...
+
+    docker-compose -f "$composeFilePath"/docker-compose.yml up -d
 elif [ "$1" == "down" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml stop
+    docker-compose -f "$composeFilePath"/docker-compose-mongo.yml -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml stop
 elif [ "$1" == "destroy" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml down -v
+    docker-compose -f "$composeFilePath"/docker-compose-mongo.yml -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml down -v
 else
     echo "Valid options are: up, down, or destroy"
```
MattyJ007, line 22: @bradsawadye This `Valid options` set needs to be updated for all the relevant bash scripts

MattyJ007: @bradsawadye Have you addressed this?

MattyJ007:
@bradsawadye You can also delete the mapper-endpoints sub-directory of the importer directory.  https://github.com/openhie/instant/tree/master/healthworkforce/docker/mapper-endpoints

healthworkforce/mock-services/endpoint.json
```diff
@@ -0,0 +1,103 @@
+{
```
MattyJ007, line 1: @bradsawadye Why is this file duplicated? It already exists in the Importer directory

healthworkforce/docker/importer/volume/endpoint.json
```diff
@@ -0,0 +1,103 @@
...
+          "method": "get",
+          "url": "http://mock-services:4000/gofr-location-mock/_history",
+          "headers": {
+            "Content-Type": "application/json"
+          },
+          "params": {
+            "_since": {
+              "path": "state.system.timestamps.lookupRequests.gofr-location.requestStart",
+              "prefix": null,
+              "postfix": null
```
MattyJ007, line 24: There are trailing blank spaces on all of the postfix fields.

```suggestion
              "postfix": null
```

core/docker/compose.sh
bradsawadye: Hey man, so on this, the mongo script is not on that location. Its in the core component. The destroy command in the core compose script handles this

MattyJ007: @bradsawadye 
```suggestion
    echo "Valid options are: init, up, down, or destroy"
```

healthworkforce/mock-services/README.md
```diff
@@ -0,0 +1,43 @@
+# mCSD endpoint flow
+
+This guide will assist in getting the mCSD endpoint setup and tested with the mock service to return the relevant lookup results
+
+## Create Endpoint record
+
+```sh
+curl -X POST -d "@endpoint.json" -H "Content-Type: application/json" http://localhost:3003/endpoints
```
MattyJ007, line 8: @bradsawadye After removing the duplicate endpoint.json file this reference needs to point to the remaining file.

healthworkforce/docker/compose.sh
```diff
@@ -1,7 +1,9 @@
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
-if [ "$1" == "up" ]; then
-    docker-compose -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml  up -d
+if [ "$1" == "init" ]; then
```
MattyJ007, line 3: @bradsawadye add `init` to echoed options below

healthworkforce/mock-services/README.md
MattyJ007: @bradsawadye

instant.sh
```diff
@@ -9,7 +9,10 @@ if ! [[ "$TARGET" =~ ^(docker|kubernetes|k8s)$ ]]; then
 fi
 
 if [ "$TARGET" == "docker" ]; then
-    if [ "$COMMAND" == "up" ]; then
+    if [ "$COMMAND" == "init" ]; then
```
MattyJ007, line 12: @bradsawadye add `init` as valid option in echo below

healthworkforce/docker/importer/volume/endpoint.json
```diff
@@ -0,0 +1,103 @@
...
+          "method": "get",
+          "url": "http://mock-services:4000/ihris-practitionerRole-mock/_history",
+          "headers": {
+            "Content-Type": "application/json"
+          },
+          "params": {
+            "_since": {
+              "path": "state.system.timestamps.lookupRequests.ihris-practitionerRole.requestStart",
+              "prefix": null,
+              "postfix": null
```
MattyJ007, line 75: @bradsawadye as I said in my previous comment, there are trailing spaces on **every postfix** field in this file... Not just one. 

```suggestion
              "postfix": null
```

PR #42, opened 2020-05-13
Ohie 388 create mscd mediator channel on start up

bradsawadye:


healthworkforce/docker/importer/docker-compose.config.yml
```diff
@@ -11,11 +11,13 @@ services:
...
+      MEDIATOR_API_PORT: 3003
       # Reject unauthorised is only needed if the OpenHIM's SSL is not setup
       NODE_TLS_REJECT_UNAUTHORIZED: 0
     volumes:
       - type: bind
         source: ./importer/volume
         target: /importer
     # This command will only attempt to import the OpenHIM config when the heartbeat responds with a 2xx
-    command: sh -c "wait-on -t 60000 https-get://core:8080/heartbeat && node openhimConfig.js"
+    command: sh -c "wait-on -t 60000 https-get://core:8080/heartbeat && node openhimConfig.js; wait-on -t 60000 http-get://mcsdMediator:3003/endpoints && node endpoint.js"
```
MattyJ007, line 23: @bradsawadye I think that this command here is dealing with two separate concerns i.e. openhim config and mediator config. It would be better to not mix our concerns and instead add a second config importer container to handle the mapping mediator config. This chaining of commands would work but it could create a bad precedent that all config can be shoved into one container.
The services could perhaps look something like this:
```yml
  openhim-config-importer:
    container_name: openhim-config-importer
    image: jembi/instantohie-config-importer
    restart: on-failure
    environment:
      OPENHIM_API_HOSTNAME: 'openhim-core'
      OPENHIM_API_PORT: 8080
      OPENHIM_API_USERNAME: 'root@openhim.org'
      OPENHIM_API_PASSWORD: 'instant101'
      # Reject unauthorised is only needed if the OpenHIM's SSL is not setup
      NODE_TLS_REJECT_UNAUTHORIZED: 0
    volumes:
      - type: bind
        source: ./importer/volume
        target: /importer
    # This command will only attempt to import the OpenHIM config when the heartbeat responds with a 2xx
    command: sh -c "wait-on -t 60000 https-get://core:8080/heartbeat && node openhimConfig.js"

  mcsd-config-importer:
    container_name: mcsd-config-importer
    image: jembi/instantohie-config-importer
    restart: on-failure
    environment:
      MEDIATOR_HOSTNAME: 'mcsdMediator'
      MEDIATOR_API_PORT: 3003
    volumes:
      - type: bind
        source: ./importer/volume
        target: /importer
    # This command will only attempt to import the mcsd mediator config when the uptime endpoint responds with 200
    command: sh -c "wait-on -t 60000 http-get://mcsdMediator:3003/uptime && node endpoint.js"

```

bradsawadye: addressed in https://github.com/openhie/instant/pull/42/commits/c2afed4e0ec5fbbf28058d34d6234ee01c650f2b

PR #44, opened 2020-05-21
Ohie 395 k8s mongo replica

bradsawadye:


core/kubernetes/main/mongo/mongo-service.yaml
```diff
@@ -1,12 +1,13 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: openhim-mongo-service
+  name: mongo
```
BMartinos, line 4: Lets keep the naming of the service with a postfix of `-service` so that its the same as the other services. If this is redundant then we should remove it from the other as well, but maybe lets keep it for now and can update at a later stage

core/kubernetes/main/mongo/initiateReplicaSet.sh
```diff
@@ -0,0 +1,18 @@
...
+
+config='{"_id":"mongo-set","members":[{"_id":0,"priority":1,"host":"mongo-0.mongo:27017"},
+{"_id":1,"priority":0.5,"host":"mongo-1.mongo:27017"},{"_id":2,"priority":0.5,"host":"mongo-2.mongo:27017"}]}'
+
+# Sleep to ensure all the mongo instances for the replica set are up and running
+sleep 20
+
+kubectl exec -i mongo-0 -n core-package -- mongo --eval "rs.initiate($config)"
+
+## Allow the replica members to stable
```
BMartinos, line 15: Not sure what is meant with `... to stable`? Unless this is just a typo :)

core/kubernetes/main/mongo/mongo-service.yaml
bradsawadye: addressed https://github.com/openhie/instant/pull/44/commits/461f957fada1c5f1e1a1d02cc9281189903b128b

core/kubernetes/main/mongo/initiateReplicaSet.sh
bradsawadye: addressed in https://github.com/openhie/instant/pull/44/commits/e81d568b38a12c6eed993d64fcc2ee9a6fb02399

PR #48, opened 2020-07-13
OHIE-420 docker outside of docker

rcrichton:


citizenrich:
Thanks for this! Looks great.

docs/docs/introduction/getting-started.mdx
```diff
@@ -64,54 +97,64 @@ To start all the default Instant OpenHIE packages, select a deployment platform
...
 
-**Before proceeding**, make sure you are in the main `/instant` directory. This directory contains the `instant.sh` script.
+If you have the Instant [OpenHIE codebase](https://github.com/openhie/instant) checked-out make sure you are in the main `/instant` directory.
 
 1. To start up the system, execute the following command:
 
     ```sh
-    ./instant.sh up docker
+    yarn
+    yarn docker:instant init
```
BMartinos, line 106: I think lets add in the specific environment we are targeting just to be a bit more verbose and inform the reader its possible to set the target environment instead of just falling back to the default

docs/docs/introduction/getting-started.mdx
```diff
@@ -64,54 +97,64 @@ To start all the default Instant OpenHIE packages, select a deployment platform
...
 
-1. To start up the system, execute the following command:
+If you have the Instant [OpenHIE codebase](https://github.com/openhie/instant) checked-out make sure you are in the main `/instant` directory.
 
-    > This command will output urls from which you can access your Instant OpenHIE instance.
+1. To start up the system, execute the following command:
 
     ```sh
-    ./instant.sh up k8s
+    yarn
```
BMartinos, line 137: Would we require the user to also run the `build` command before starting the kubernetes instance?

docs/docs/introduction/getting-started.mdx
```diff
@@ -64,54 +97,64 @@ To start all the default Instant OpenHIE packages, select a deployment platform
...
+    > **This action will also delete all volumes** created by the containers.
 
     ```sh
-    ./instant.sh destroy k8s
+    yarn docker:instant destroy -t kubernetes
     ```
 
+Each command also takes a list of package IDs to operate on. If this is left out then all packages are run by default.
+
+If you do not have the code checked-out then you can still startup the environment with the correct command. You can find the docker commands to run in the scripts section of [this file](https://github.com/openhie/instant/blob/ohie-420-docker-outside-of-docker/package.json) in the codebase. Add the action you would like to perform (init, up, down, destroy) to the end of that command and make sure to add the `-t kubernetes` option after the action.
```
BMartinos, line 157: Maybe we can update this link so long to point to the location of where it will be once in master, to avoid having another PR for correcting the file location

goinstant/goinstant.go
```diff
@@ -117,29 +124,31 @@ func composeUp(composeFile string) {
...
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 		err := cmd.Run()
 		if err != nil {
 			log.Fatalf("cmd.Run() failed with %s\n", err)
 		}
 	case "windows":
-		cmd := exec.Command("cmd", "/C", "docker-compose", "-f", "-", "down")
-		cmd.Stdin = strings.NewReader(composeFile)
+		cmd := exec.Command("cmd", "/C", "docker", "run", "--rm", "-v", "/var/run/docker.sock:/var/run/docker.sock", "-v", home+"\\.kube:/root/.kube/config:ro", "--mount=type=volume,src=instant,dst=/instant", "openhie/instant:test", "down", "-t", "docker")
```
BMartinos, line 144: I see we make use of a `openhie/instant:test` image. Is the tag we are using (`test`) correct

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:lts
+
+WORKDIR /instant
+
+# install curl
+RUN apt-get update; apt-get install -y curl
+
+# install kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
```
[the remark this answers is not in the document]
rcrichton: Hey, this wasn't my command, it is the recommended one liner to install Kubectl. While I agree with your comment, I feel it's better to align with how this command appears elsewhere on the internet.

docs/docs/how-to/creating-packages.mdx
```diff
@@ -84,7 +88,12 @@ A `k8s.sh` script could look like this:
 
 k8sMainRootFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
-if [ "$1" == "up" ]; then
+if [ "$1" == "init" ]; then
+    echo "execute other setup scripts or commands here..."
```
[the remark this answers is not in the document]
rcrichton: So this was meant to distinguish the `init` code block from the `up` code block. Maybe it should just be a comment rather than an echo command?

goinstant/goinstant.go
rcrichton: This should be latest, but that will only get built when this is in master.

docs/docs/introduction/getting-started.mdx
rcrichton: It's not required. If you don't build locally it will use the image from docker hub.

docs/docs/how-to/creating-packages.mdx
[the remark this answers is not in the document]
BMartinos: ok cool, yeah i think a comment might be better instead of an echo. Just wasnt sure what text was meant for. Maybe the comment can be a bit more descriptive to indicate its used for any additional initialization scripts for the setup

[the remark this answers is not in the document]
BMartinos: Actually, lets leave it as the comment you have :+1:

instant.ts
```diff
@@ -0,0 +1,177 @@
...
+
+function getInstantOHIEPackages(): PackagesMap {
+  const packages: PackagesMap = {}
+  const paths = glob.sync('*/instant.json')
+
+  for (const path of paths) {
+    const metadata = JSON.parse(fs.readFileSync(path).toString())
+    packages[metadata.id] = {
+      metadata,
+      path: path.replace('instant.json', ''),
```
BMartinos, line 32:
```suggestion
      path: path.replace('instant.json', '')
```

instant.ts
```diff
@@ -0,0 +1,177 @@
+import * as commandLineArgs from 'command-line-args'
```
BMartinos, line 1:
```suggestion
'use strict'

import * as commandLineArgs from 'command-line-args'
```

instant.ts
```diff
@@ -0,0 +1,177 @@
...
+      .map((p) => p.metadata.id)
+      .join(', ')}`
+  )
+
+  const main = commandLineArgs(
+    [
+      {
+        name: 'command',
+        defaultOption: true,
+      },
```
BMartinos, line 73:
```suggestion
      }
```

instant.ts
```diff
@@ -0,0 +1,177 @@
...
+    ],
+    {
+      stopAtFirstUnknown: true,
```
BMartinos, line 76:
```suggestion
      stopAtFirstUnknown: true
```

instant.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  // main commands
+  if (['init', 'up', 'down', 'destroy'].includes(main.command)) {
+    const mainOptions = commandLineArgs(
+      [
+        {
+          name: 'target',
+          alias: 't',
+          defaultValue: 'docker',
+        },
```
BMartinos, line 90:
```suggestion
        }
```

instant.ts
```diff
@@ -0,0 +1,177 @@
...
+        {
+          name: 'host',
+          alias: 'h',
+          defaultValue: 'localhost',
+        },
+        {
+          name: 'port',
+          alias: 'p',
+          defaultValue: '5000',
+        },
```
BMartinos, line 150:
```suggestion
        }
```

PR #50, opened 2020-07-15
OHIE-422 Core Tests

MattyJ007:
Share partially working example.

The Gherkin keyword `Example` in the ticket definition didn' work???
Replaced with Scenario.


core/features/core.js
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    const response = await axios(options)
+    expect(response.status).to.eql(201)
+  } else {
+    console.log(`The Test Harness Client (Alice) already exists...`)
+  }
+})
+
+Given('a patient, Jane Doe, exists in the FHIR server', async function () {
```
rcrichton, line 125: @MattyJ007 the intention of the given statements is to give you an opportunity to create the needed test data, so I think you can move the test data creation fro the before all to the given section. This should reduce the amount of checking that you need to do as well.

rcrichton: > When Cucumber executes a Given step, it will configure the system to be in a well-defined state, such as creating and configuring objects or adding data to a test database.

https://cucumber.io/docs/gherkin/reference/

rcrichton:
Functional testing went well. All passed.

core/features/core.js
MattyJ007: Did this to avoid duplicate creation before I knew it was HAPI causing it :| 

Ge-sorted :+1:

PR #52, opened 2020-07-22
OHIE-425 Run cucumber from instant.ts script

de-laz:


core/docker/compose.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
rcrichton, line 1: Looks like the permissions got changed from 755 to 644 on this file. I think that needs to be reverted. Seems like the execute flag got removed.

PR #53, opened 2020-07-23
Ohie 423 mcsd package feature

bradsawadye:


healthworkforce/features/utils.js
```diff
@@ -0,0 +1,283 @@
...
+  await axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/mcsd-trigger`,
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Basic ${authHeader}`
+    }
+  })
+  .then(() => {
+    console.log('Data successfully sent to dhis')
```
rcrichton, line 36: This mentions DHIS2

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,283 @@
...
+  })
+  .catch((err) => {
+    console.log('Error in sending data to dhis', err.message)
```
rcrichton, line 39: This mentions DHIS2

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,283 @@
...
+  })
+}
+
+exports.ihrisMockServicePractitioner = async () => {
+  await verifyResourceDoesNotExist('Practitioner')
+
+  nock('http://mock-service:4000')
```
rcrichton, line 46: Like Martin was saying in the stand up, instead of using nock we should rather allow the mock service to reply to this request. Even if it just gives a simple static response.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,283 @@
...
+exports.verifyPractitionerExistsAndCleanup = async () => {
+  const resource = 'Practitioner'
+  const response = await getResource(resource)
+
+  expect(response.status).to.eql(200)
+  expect(response.data.name[0].given, practitionerName)
+
+  await removeResource(resource)
+  .then(() => {console.log(`Test ${resource} successfully removed from FHIR`)})
+  .catch(() => {console.error(`Clean up failed, test ${resource} not removed from FHIR`)})
```
rcrichton, line 246: If you are using `await` then you don't need to use `.then` and `.catch`. `await` will pause until it completes so any statement written after this is the same as doing a `.then`. For `.catch` you should wrap the entire block in a `try {} catch ()` statement.

healthworkforce/features/utils.js
rcrichton: You might have to add the test resources to the data that the mock service is already returning so that the test still function. However, this is better because at least when the real services are available all the code stays the same and we just change the urls.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,283 @@
...
+const locationName = "GoodHealth Clinic"
+const organizationName = "Clinical Lab"
+
+exports.triggerSync = async () => {
+  await axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/mcsd-trigger`,
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Basic ${authHeader}`
```
rcrichton, line 32: I don't think we need this authHeader as the channels are public right now.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+  await verifyResourceDoesNotExistInFHIR('Organization', organizationId)
+}
+
+const getResource = (resource, id) => {
+  return axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/hapi-fhir-jpaserver/fhir/${resource}?_id=${id}`,
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Custom ${CUSTOM_TOKEN_ID}`
```
MattyJ007, line 57:
```suggestion
      Authorization: `Custom ${CUSTOM_TOKEN_ID}`,
      'Cache-Control': 'no-cache'
```

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+  })
+}
+
+const removeResource = (resource, id) => {
+  return axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/hapi-fhir-jpaserver/fhir/${resource}?_id=${id}`,
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Custom ${CUSTOM_TOKEN_ID}`
```
MattyJ007, line 68:
```suggestion
      Authorization: `Custom ${CUSTOM_TOKEN_ID}`,
      'Cache-Control': 'no-cache'
```

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+const OPENHIM_PROTOCOL = process.env.OPENHIM_PROTOCOL || 'http'
+const OPENHIM_API_HOSTNAME = process.env.OPENHIM_API_HOSTNAME || 'localhost'
+const OPENHIM_TRANSACTION_API_PORT =
+  process.env.OPENHIM_TRANSACTION_API_PORT || '5001'
+const CUSTOM_TOKEN_ID = process.env.CUSTOM_TOKEN_ID || 'test'
+
+const practitionerId = 'P10004'
+const practitionerRoleId = 'PR10001'
+const locationId = '2'
+const organizationId = '3'
```
rcrichton, line 14: @bradsawadye You can't use the hard coded IDs as these will change depending on what already exists in the database. Perhaps rather add a test `identifier` that you can search for.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/mcsd-trigger`,
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json'
+    }
+  })
+
+  // Allow the syncing to finish. The data will only be available in FHIR after a few seconds
+  await new Promise ((resolve, _reject) => {
+    setTimeout(() => resolve(), 60000)
```
rcrichton, line 27: We could probably change this to be something smaller. Maybe optimise it once all is completed.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+}
+
+exports.gofrMockServiceOrganization = async () => {
+  // Mock gofr service running in docker container
+  await verifyResourceDoesNotExistInFHIR('Organization', organizationId)
+}
+
+const getResource = (resource, id) => {
+  return axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/hapi-fhir-jpaserver/fhir/${resource}?_id=${id}`,
```
rcrichton, line 53: The more usual way to get a resource by `id` is like this `/hapi-fhir-jpaserver/fhir/${resource}/${id}`

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Custom ${CUSTOM_TOKEN_ID}`
+    }
+  })
+}
+
+const removeResource = (resource, id) => {
+  return axios({
+    url: `${OPENHIM_PROTOCOL}://${OPENHIM_API_HOSTNAME}:${OPENHIM_TRANSACTION_API_PORT}/hapi-fhir-jpaserver/fhir/${resource}?_id=${id}`,
```
rcrichton, line 65: The more usual way to delete a resource by `id` is like this `/hapi-fhir-jpaserver/fhir/${resource}/${id}`

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,151 @@
...
+  })
+
+  // Allow the syncing to finish. The data will only be available in FHIR after a few seconds
+  await new Promise ((resolve, _reject) => {
+    setTimeout(() => resolve(), 60000)
+  })
+}
+
+exports.ihrisMockServicePractitioner = async () => {
+  // Mock ihris service running in docker container
```
rcrichton, line 32: For this step definition we still need the code that creates the FHIR resource - we should POST a FHIR resource that the mock service is already returning (or modify one that it's returning to match). The mock service can just accept the create request and do nothing with it. When we finally do have actual services, then all should just work when we remove the mock services.

healthworkforce/features/utils.js
bradsawadye: addressed

healthworkforce/features/utils.js
bradsawadye: addressed

healthworkforce/features/utils.js
bradsawadye: addressed

healthworkforce/features/utils.js
bradsawadye: addressed

healthworkforce/features/utils.js
bradsawadye: addressed

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,327 @@
...
+  if (response.data.total > 0) throw Error(
+    `Test aborted! ${resource} resource (identifier-value: ${resourceId}) used in test already exists and will be removed from the FHIR server`
+    )
+}
+
+const createResourceBundle = (resourceName, resourceBundle) => {
+  return axios({
+    url: `http://localhost:${MOCK_SERVER_PORT}/create-resource/${resourceName}`,
+    method: 'POST',
+    data: resourceBundle
```
rcrichton, line 211: We should use the normal FHIR resource creation here as that is what we will be using with the actual (non-mock) services. This means we should send the resource itself instead of a bundle to our mock services. I think all you need to change is the post bodies to contain just the resource.

healthworkforce/features/utils.js
```diff
@@ -0,0 +1,327 @@
...
+
+  // Remove resource from FHIR
+  const result = await removeResource(resource, identifierValue)
+
+  if (
+    !JSON.stringify(result.data).match(/Successfully deleted 1 resource/)
+    ) throw(`Clean up failed, test ${resource} (identifier-value: ${identifierValue}) not removed from FHIR`)
+
+  // Remove the practitionerRole bundle created on the ihris server
+  const result_2 = await deleteResourceBundle(resource)
```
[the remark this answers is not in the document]
rcrichton: Maybe rename the result above to make it explicit to?

PR #63, opened 2020-12-04
order the packages

bradsawadye:
Packages that other packages are dependent on should be started first

DSC19-7

### DIFF
--- a/instant.ts
+++ b/instant.ts
@@ -75,85 +75,34 @@ async function runTests(path: string) {
 }
 
 const orderPackageIds = (allPackages, chosenPackageIds) => {
-  const orderedPackageIds = []
-  const packagesWithDependencies = []
+  function resolveDeps(id, currentStack) {
+    if (currentStack.includes(id)) throw Error(`Circular dependency present for id ${id}`)
+    currentStack.push(id)
 
-  function collectPackagesWithDependencies(chosenIds) {
-    let inexplicitDependencies = []
-
-    chosenIds.forEach(id => {
-      if (id && allPackages[id] && allPackages[id].metadata) {
-        if (
-          !allPackages[id].metadata.dependencies ||
-          !allPackages[id].metadata.dependencies.length
-        ) {
-          if (id === 'core') {
-            orderedPackageIds.unshift(id)
-          } else {
-            orderedPackageIds.push(id)
-          }
-        } else {
-          allPackages[id].metadata.dependencies.forEach(dependency => {
-            if (
-              !Object.keys(allPackages).includes(dependency) ||
-              !allPackages[dependency].metadata
-            ) throw Error(
-              `Dependency ${dependency} for package ${id} does not exist or the metadata is invalid`
-            )
-            if (
-              allPackages[dependency].metadata.dependencies &&
-              allPackages[dependency].metadata.dependencies.length &&
-              allPackages[dependency].metadata.dependencies.includes(id)
-            ) throw Error(
-              `Circular dependencies - ${id} and ${dependency}`
-            )
-
-            if (!chosenIds.includes(dependency)) {
-              inexplicitDependencies.push(dependency)
-            }
-          })
-          packagesWithDependencies.push(id)
-        }
-      } else {
-        throw Error(`Package ${id} does not exist or the metadata is invalid`)
-      }
-    })
-
-    if (inexplicitDependencies.length) {
-      collectPackagesWithDependencies(inexplicitDependencies)
+    if (allPackages[id] && allPackages[id].metadata) {
+      if (
+        !allPackages[id].metadata.dependencies ||
+        !allPackages[id].metadata.dependencies.length
+      ) return [id]
     } else {
-      return
+      throw Error(`Package ${id} does not exist or the metadata is invalid`)
     }
+
+    let orderedIds = []
+
+    allPackages[id].metadata.dependencies.forEach(dependency => {
+      const ids = resolveDeps(dependency, currentStack)
+      currentStack = [id]
+      orderedIds.push(...[...ids, id])
+    })
+    return orderedIds
   }
 
-  collectPackagesWithDependencies(chosenPackageIds)
-
-  while (packagesWithDependencies.length) {
-    const currentPackagesLength = packagesWithDependencies.length
-
-    for (let index = 0; index < packagesWithDependencies.length; index++) {
-      const id = packagesWithDependencies[index]
-      let containDependencies = true
-
-      allPackages[id].metadata.dependencies.forEach(dependency => {
-        if (!orderedPackageIds.includes(dependency)) {
-          containDependencies = false
-        }
-      })
-
-      if (containDependencies) {
-        orderedPackageIds.push(id)
-        packagesWithDependencies.splice(index, 1)
-      }
-    }
-    /*
-      If circular dependencies are present, the array of packages with dependencies will
-      not change after the loop above finishes
-    */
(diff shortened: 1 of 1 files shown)


instant.ts
```diff
@@ -74,6 +74,46 @@ async function runTests(path: string) {
...
+      allPackages[id].metadata.dependencies.forEach(dependency => {
+        if (!Object.keys(allPackages).includes(dependency)) {
+          throw Error(`Dependency ${dependency} for package ${id} does not exist`)
+        }
+      })
+      packagesWithDependencies.push(id)
+    }
+  })
+
+  while (packagesWithDependencies.length) {
```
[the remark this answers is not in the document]
rcrichton: This is quite a brute force way of doing things. You might struggle to be able to detect circular dependencies with this method.

An alternative method that you can consider is to try find a javascript library that support a tree data structure. Then you could load all the packages into a tree based on which packages are dependant on others. Then would could traverse the tree to get the order in which they must start up. A tree should also allow you to detect circular dependencies.

Maybe spend an hour looking for a library that is well used and supports tree traversal and detecting circular references. If there is a good one then maybe we can switch to using that. If you can't find anything, then we can just use what you have here and add a comment to come back to this and add circular dependency handling if it becomes a big problem. We don't want to invest too much time into this if it proves to be difficult.

rcrichton (changes requested):
@bradsawadye thanks, I've added some comment for you to consider.

instant.ts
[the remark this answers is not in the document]
bradsawadye: @rcrichton, I could not find a good library for dealing with this. I've added some logic to cater for cases when we have circular dependencies

instant.ts
```diff
@@ -74,6 +74,59 @@ async function runTests(path: string) {
   }
 }
 
+const orderPackageIds = (allPackages, chosenPackageIds) => {
```
rcrichton, line 77: If the `chosenPackageIds` have dependencies that aren't explicitly lsited in the `chosenPackageIds` then we should load those into the list. This way we can choose a single package and the script can ensure that all its dependencies are added in.

instant.ts
```diff
@@ -127,6 +180,9 @@ async function runTests(path: string) {
       chosenPackageIds = Object.keys(allPackages)
     }
```
rcrichton, line 182: When running a `stop` or `destroy` we need to use the reverse order so dependencies are stopped before the dependant packages.

instant.ts
```diff
@@ -74,6 +74,59 @@ async function runTests(path: string) {
...
+      if (containDependencies) {
+        orderedPackageIds.push(id)
+        packagesWithDependencies.splice(index, 1)
+      }
+    }
+    /*
+      If circular dependencies are present, the array of packages with dependencies will
+      not change after the loop above finishes
+    */
+    if (currentPackagesLength == packagesWithDependencies.length) {
```
rcrichton, line 123: Good work! That's a pretty clever solution.

instant.ts
bradsawadye: addressed https://github.com/openhie/instant/pull/63/commits/725f533b6177a0839464ada7d70db41cd6513dbe

instant.ts
bradsawadye: addressed in https://github.com/openhie/instant/pull/63/commits/1083e15e0136b5525bc7033b608abdc4c6f3a3df

instant.ts
```diff
@@ -78,23 +78,48 @@ const orderPackageIds = (allPackages, chosenPackageIds) => {
...
         }
-      })
-      packagesWithDependencies.push(id)
+      } else {
+        throw Error(`Package ${id} does not exist or the metadata is invalid`)
+      }
+    })
+
+    if (inexplicitDependencies.length) {
+      collectPackagesWithDependencies(inexplicitDependencies)
```
rcrichton, line 112: This recursive function needs protection against circular dependencies else it looks until `RangeError: Maximum call stack size exceeded`

rcrichton (changes requested):
@brad looking good just one more comment.

instant.ts
bradsawadye: @rcrichton addressed in https://github.com/openhie/instant/pull/63/commits/5a7840d5f0a619966b6639e6705598c7ff35dae6

instant.ts
```diff
@@ -74,6 +74,89 @@ async function runTests(path: string) {
...
+            if (
+              !Object.keys(allPackages).includes(dependency) ||
+              !allPackages[dependency].metadata
+            ) throw Error(
+              `Dependency ${dependency} for package ${id} does not exist or the metadata is invalid`
+            )
+            if (
+              allPackages[dependency].metadata.dependencies &&
+              allPackages[dependency].metadata.dependencies.length &&
+              allPackages[dependency].metadata.dependencies.includes(id)
```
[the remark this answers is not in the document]
rcrichton: Thinking about this recursive function. I wonder if you could expand the inexplict dependencies and do the ordering like in the `while` loop below in a single recursive function as you are already having to go through every dependency. This might simplify this alot. Something like:

```
resolveDeps(id, currentStack) {
  if currentStack contains id
    throw circular dep error
  else push id to currentStack

  if id has no deps
    return [id]

  ordferedIds = []
  for each dep of id
    orderedIds = resolveDeps(dep, currentStack)
    orderedIds.push(dep)

  return orderedIds
}

finalOrderedPackages = []
foreach chosenId
  finalOrderedPackages.push(resolveDeps(chosenId, []))

// at this point you may have to remove any duplicate package Ids from the array.
```

rcrichton (changes requested):
This still doesn't work for every case. I've added some comments that might help.

instant.ts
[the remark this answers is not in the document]
bradsawadye: good solution you have here. I've addressed all in https://github.com/openhie/instant/pull/63/commits/f983bc037e991eb4713fa96a35e56d1e8e680e35

instant.ts
```diff
@@ -74,6 +74,38 @@ async function runTests(path: string) {
...
+        !allPackages[id].metadata.dependencies.length
+      ) return [id]
+    } else {
+      throw Error(`Package ${id} does not exist or the metadata is invalid`)
+    }
+
+    let orderedIds = []
+
+    allPackages[id].metadata.dependencies.forEach(dependency => {
+      const ids = resolveDeps(dependency, currentStack)
```
rcrichton, line 94: Just an improvement: I think here we should rather create a copy of the `currentStack` each time the loop runs. This way when the current stack changes it won't affect other loops and we don't have to reset it.

instant.ts
```diff
@@ -74,6 +74,38 @@ async function runTests(path: string) {
...
+      currentStack = [id]
```
rcrichton, line 95: Then, this line can go away as well.

instant.ts
```diff
@@ -74,6 +74,38 @@ async function runTests(path: string) {
...
+      orderedIds.push(...[...ids, id])
```
rcrichton, line 96: I'm seeing an error in testing, I think it's because the `id` needs to only be added after the loop completes.

rcrichton (changes requested):
@bradsawadye Looking good, I added a few comment for things that I found.

instant.ts
bradsawadye: addressed in https://github.com/openhie/instant/pull/63/commits/53bfe709459405ef33b3da3acc088062c7aef375

rcrichton (approved):
Awesome! Looks good to me now.

PR #72, opened 2021-03-04
DEB-5 verify opencr

MattyJ007:
Add volumed config to Client package


citizenrich (commented):
Looks good. Though I'd prefer env vars for the entire file, I wonder if it makes sense to have `config_instant for OpenCR live inside Instant or if it can be dropped into OpenCR if it won't change. NODE_ENV should be able to invoke it if we put it in directly.

rcrichton:
@citizenrich I think adding in env vars for the all the changes in the config and matching files might be too much. It would be better if OpenCR allowed an env var to specify where the config directory was, then we wouldn't need to go through all of this. Maybe that is something you would like to get on the roadmap, it should be a simple change.

Also, I don't think the changes Matt made have been built on Dockerhub yet for OpenCR, if you could push to get those up it would be awesome.

client/docker/importer/volume/config_instant.json
```diff
@@ -0,0 +1,79 @@
...
+  "clients": [
+    {
+      "id": "openmrs",
+      "displayName": "OpenMRS"
+    },
+    {
+      "id": "lims",
+      "displayName": "Lab Info Management System"
+    }
+  ],
```
MattyJ007, line 51: @citizenrich Should we remove these clients? They don't seem related.
Do they need to be replaced with an Instant client?

client/docker/importer/volume/config_instant.json
```diff
@@ -0,0 +1,79 @@
...
+      "uri": "http://health.go.ug/cr/natioanlid",
+      "displayName": "National ID"
+    },
+    "artnumber": {
+      "uri": "http://health.go.ug/cr/artnumber",
+      "displayName": "ART Number"
+    },
+    "brokenMatch": {
+      "uri": "http://clientregistry.org/brokenMatch"
+    }
```
MattyJ007, line 71: @citizenrich Can we leave these systems out? Or make them more generic?

MattyJ007 (commented):
@rcrichton Thoughts?

citizenrich:
> @citizenrich I think adding in env vars for the all the changes in the config and matching files might be too much. It would be better if OpenCR allowed an env var to specify where the config directory was, then we wouldn't need to go through all of this. Maybe that is something you would like to get on the roadmap, it should be a simple change.
> 
> Also, I don't think the changes Matt made have been built on Dockerhub yet for OpenCR, if you could push to get those up it would be awesome.

Pushed.

citizenrich:
> @citizenrich I think adding in env vars for the all the changes in the config and matching files might be too much. It would be better if OpenCR allowed an env var to specify where the config directory was, then we wouldn't need to go through all of this. Maybe that is something you would like to get on the roadmap, it should be a simple change.

@ashaban Can you have a look at adding an env var in OpenCR for the location of the config file directory? The default is can still be /server/config but having this feature would help out with cleaner use of OpenCR in Instant.

client/docker/importer/volume/config_instant.json
citizenrich: @ashaban Why do we enumerate clients in the config file?

MattyJ007:
@citizenrich From our meeting, earlier this week I'm now more hesitant to use Environment Variables in OpenCR to configure anything custom package specific. This is due to the fact that to change Environment variables - i.e. to point to a new set of config files - you would need to destroy the existing OpenCR docker container and create a new one. This would entail adding new OpenCR docker-compose scripts into custom packages - this would have a high maintenance cost and is mixing up of package concerns. I think custom packages should only have config scripts for core package components and the bare minimum (preferably nothing) for core package infrastructure set up.
Using this volume method, when a new OpenCR config is needed by a custom package we only require restarting the container to apply changes. I think this is safer and easier for custom package implementors who won't necessarily have the context of the core Instant pacakges.
@rcrichton Thoughts?

ashaban:
Im thinking of moving some of these config to a fhir resource and save them
to the db, and allow people to change these resources via environment
variables, will this help?

MattyJ007:
@ashaban I don't think the OpenCR config should go in HAPI-FHIR as that is not what HAPI-FHIR is designed for - you would probably need to hack a FHIR Resource to store that kind of data as no resource applies here as far as I know.

Also, the only config that I foresee needing updating often is the `systems` uri and names section. As we discussed in the meeting earlier in the week there may be a way to get around that - perhaps we can organise another meeting to discuss this feature at some point once we've looked at it a bit deeper

ashaban:
@MattyJ007 we can use the Parameters resource for configurations, we can as well use the Device resource to keep track details about devices (clients) that submits patients to openCR, with the Device resource, we can move config:clients, config:systems:internalid:uri to the the Device resource as follows
{
  "resourceType": "Device",
  "distinctIdentifier": "http://openmrs.org/openmrs2",
  "deviceName": [{
    "name": "OpenMRS",
    "type": "user-friendly-name"
  }]
}

citizenrich:
+1 to use FHIR resources inside HAPI where it makes sense. HAPI (SmileCDR team) uses a combination themselves, of a custom matching rules JSON file and a FHIR Parameter resource for links.

Footnote: IMHO, at some point in the future, some users could choose to use OpenCR as just a front-end to the HAPI MDM engine, which means anyone using HAPI has an easy NodeJS-based CR to use. This would mean easier scale for OpenCR by the HAPI community. So, if some use cases requires a simple, deterministic CR, they can use HAPI + OpenCR UI. If they want ElasticSearch etc then they can use the current OpenCR. Obviously, make OpenCR talk to HAPI MDM requires a substantial effort. So just noting the idea here.

Another footnote: HAPI's MDM can match any resource. This means locations, practitioners, whatever record linkage needs to happen.

rcrichton:
Use a pinned version of OpenCR


```diff
diff --git a/client/docker/docker-compose.yml b/client/docker/docker-compose.yml
index ffd398d..bc02aaa 100644
--- a/client/docker/docker-compose.yml
+++ b/client/docker/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3.3'
 services:
   opencr:
     container_name: opencr
-    image: intrahealth/opencr:latest
+    image: intrahealth/opencr:9628aa9
     # change to 3004 to avoid port contention
     ports:
       - "3004:3000"

```

rcrichton:
I'm going to merge this for now, the client package is now working. We can roll forward with more updates if necessary.

ashaban:
@MattyJ007 @rcrichton @citizenrich i have updated openCR to use a common URI for source system internal ID(which is http://openclientregistry.org/fhir/sourceid) for all all clients submitting Patients to openCR, this means we will no longer need to define internal ID URI for every client on the config and instead all clients should include their internal ID using the default system http://openclientregistry.org/fhir/sourceid
this update will be available tomorrow.

rcrichton:
Ok, thanks Ally.

citizenrich:
Pushed a new image.

PR #80, opened 2021-04-16
fix path to yml in core/docker/readme

mrjones-plip:
Based on what I see in `compose.sh` ([see line 11](https://github.com/openhie/instant/blob/master/core/docker/compose.sh#L11)), I believe the path to `docker-compose.config.yml` needs to be fixed per this PR.

### DIFF
--- a/core/docker/README.md
+++ b/core/docker/README.md
@@ -52,7 +52,7 @@ Additional `docker-compose` files are available for extra environment configurat
 The below command specifies the three `docker-compose` files that need to be executed for the development configuration
 
 ```bash
-docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose.config.yml up -d
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f importer/docker-compose.config.yml up -d
 ```
 
 The below command specifies the two `docker-compose` files that need to be executed for a production-like configuration


mrjones-plip:
In doing some testing following the same [docker compose steps](https://github.com/openhie/instant/tree/master/core/docker#prerequisites), it looks like these same instructions skip calling the `configure-docker.sh` prep script.  Further, because the instructions don't use the `compose.sh` script in the directory, the `mongodb` databases fail to to start in time and the `core` container will fail to start.

I'd be happy to update this PR with the steps on using `compose.sh` as I've just finished doing this for [our test build out](https://github.com/medic/chis-interoperability/#install)!

citizenrich:
Thanks for this PR. I'm rebuilding the CLI and see that the HAPI channel for OpenHIM is not being loaded and the importer not firing up could be the reason. I'll try this but also interested in other folks weighing in.

mrjones-plip:
> Thanks for this PR. I'm rebuilding the CLI and see that the HAPI channel for OpenHIM is not being loaded and the importer not firing up could be the reason. I'll try this but also interested in other folks weighing in.

@citizenrich  - cool! I'm relatively new to docker, and even newer to OpenHIM, so I'm happy to help but very open to other folks weighing in.

> Related: So, yeah, `init` needs to be called then `up`. I'm curious if the ergonomics make sense for users. Perhaps in future `init`, `up`, `down`, `destroy` could be replaced with start (combo of init, up), stop (stop but keep configs, volumes etc), and stop/cleanup (stop and destroy).

My reading of the `console.sh` is that `init` does everything that `up` does, but adds in  calling `initiateReplicaSet.sh`.  This means you don't need to call `up` after `init`, only after `down` has been called and you want to start the services.

I'm otherwise OK with keeping things as is and not move to `start` etc per your suggestion!  As the new kid on the block, feel free to ignore though ;)

Speaking of related: I've been running the change in this PR against our project and I think I'm seeing the `core-config-importer` container repeatedly run, such that our passwords for `test` Client and `root@openhim.org` User keep on getting reset.  I'm not sure if this is a feature of the dev environment, such that you intentionally regularly reset the environment, or if this is a bug such that I should open another ticket (or if my environment is special and it's somehow  neither of these!)

danfuterman:
@mrjones-plip Thanks for this PR and the other PR/issues logged. We've got some time scheduled in our next sprint to look into these issues and PRs, and will reach out if we have any questions/suggestions.

MattyJ007 (approved):
LGTM

Definitely a mistake in the file path

mrjones-plip:
@MattyJ007  - awesome - glad this PR made it in!

Issue #79, opened 2021-04-16
yarn init calls docker compose differently than yarn up


mrjones-plip:
Hi there!  I followed  [the instructions to use docker](https://github.com/openhie/instant#docker-compose) for initializing Instant OpenHIE as well as how to bring it down and then back up again.  I found a bug in that `init` appears to make a different call than `up` does.

Steps to reproduce:
1. on an ubuntu 18 machine, install docker and docker compose and then clone this repo
2. cd into the repo and call `yarn docker:instant init -t docker core`
3. Run `docker ps`.  Note that, for example, `hapi-fhir` container is listening on all interfaces and forwarding port `3447` to port `8080`: `0.0.0.0:3447->8080/tcp`
4. run `yarn docker:instant down -t docker core` followed by `yarn docker:instant up -t docker core`

**Expected:** All containers are the same between `init` and `up`
**Actual:** looking at  `docker ps`, note that now, for example, `hapi-fhir` container is listening on no external interfaces and not forwarding any ports: `8080/tcp`

**Workaround:** [It looks like](https://github.com/openhie/instant/blob/master/core/docker/README.md) Instead of using `yarn` and the bootstrap `openhie/instant` container, you can call `docker-compose` directly, avoiding this issue.

**Fix:** My testing suggests publishing a new [openhie/instant](https://hub.docker.com/r/openhie/instant) image with the contents of [compose.sh](https://github.com/openhie/instant/blob/master/core/docker/compose.sh) updated such that line 18 matches line 11, it should fix it.  The issue is that `init`'s call has all the right `.yml`'s included:

```bash
docker-compose -p instant -f "$composeFilePath"/docker-compose.yml -f "$composeFilePath"/docker-compose.dev.yml -f "$composeFilePath"/importer/docker-compose.config.yml up -d
```

But when you call `up`, it makes a different call with just one `yml`:

```bash
docker-compose -p instant -f "$composeFilePath"/docker-compose.yml up -d
```

MattyJ007:
Thanks @mrjones-plip fixed

PR #82, opened 2021-04-19
fix bash wrapper for docker compose 

mrjones-plip:
This PR fixes the bash script per #79.  The [docker image](https://hub.docker.com/r/openhie/instant) will still need to be rebuilt and published.


mariusnaude:
Hi @mrjones-plip , I'm with the Jembi team (analysis/project facilitation). Just want to let you know that I also experienced the same issue and our team will be picking up this PR soon!

MattyJ007 (approved):
LGMT - I just removed the config file from the up command. That should not be there

PR #85, opened 2021-04-29
Up the openhim console version minor

MattyJ007:
A useful authentication feature was added in this version. This will
also match the docker version

DSC19-35


de-laz (approved):
@MattyJ007 LGTM :+1:

PR #96, opened 2021-05-11
ClI improvements and refactor

citizenrich:
A number of issues arose with the previous version of the CLI tool. This PR does the following:
* Simplified interactive menu. The menu allows users to launch and destroy packages in Docker and Kubernetes. The backend approach to doing so is cleaner.
* Deprecated web UI in favor of the CLI.
* Adds a tool to install FHIR NPM packages to any FHIR server
* Adds the interactive CLI commands to args for CI/CD, QA etc, e.g.:
```
Commands: 
	help 		this menu
	docker		manage package in docker, usage: docker <package> <state> e.g. docker core init
	kubernetes	manage package in kubernetes, usage: k8s/kubernetes <package> <state>, e.g. k8s core init
	install		install fhir npm package on fhir server, usage: install <ig_url> <fhir_server>, e.g. install https://intrahealth.github.io/simple-hiv-ig/ http://hapi.fhir.org/baseR4
```

TODO (but maybe not in this PR)
* A way to load packages from external repos.

To test:
```
git checkout cli_refactor
cd goinstant/
go mod download
# for interactive, same as running the built executable
go run *.go
# using args
go run *.go docker core init
```


citizenrich:
Absent comments, I can merge this and tag a release but I do have an outstanding question... The yarn toolkit uses a different order for args/flags. I can revise the CLI to use the same one. The only thing I think that should be left out I think the -t flag is not necessary, as it's always the second argument. Thoughts?

rcrichton:
Sorry Richard, we need to find some time to go through this. It has been crazy with other projects. I agree about the `-t` flag, maybe that can be dropped. One thing I can quickly say (without looking at the changes you have made yet) is that we definitely need the GO script to cover all the cases that the current `deploy.sh` script does. In particular the `-c` flag to load custom packages from the local file system. Not sure if that's something you have included?

citizenrich:
Nope, that feature is missing, it used to only just load a local docker-compose. I'll have a close look at the deploy script -c flag. Thanks for the input!

citizenrich:
@michaelloosen thanks for this change. I'm revising this app at the moment and will have a go with your changes. Do you know what errors were being seen in Windows?

citizenrich:
@michaelloosen thanks for this change. I'm revising this app at the moment and will have a go with your changes. Do you know what errors were being seen in Windows?

michaelloosen:
Hi @citizenrich, I can't remember the exact error message and the underlying error, but it was giving us an exit code of 125 when executing the `err = cmd.Wait()` line in the `docker.go` file. After some research, we found that it was due to the voluming of  `/var/run/docker.sock:/var/run/docker.sock` requiring the double slash on the Windows side (which didn't make much sense). I have however just tried to recreate the issue (by removing my changes first) to see if I can encounter the exact error again and haven't been able to, so I will ask @MattyJ007 in the morning if he can try without my changes to see if he encounters it or not to determine if it's really required or if it was just a local issue.

michaelloosen:
Hi @citizenrich We've tested on multiple other Windows devices and seems the change I added wasn't actually required, so I've removed my commit. 👍

citizenrich:
Ok, thanks @michaelloosen I'm testing with Windows this week too and will have a go at it.

MattyJ007:
@citizenrich Are you happy with us merging this PR? We have a few more Cli features to add this week but for ease of review we'd like to add this work to master then create new branches

citizenrich:
@MattyJ007 Absolutely.

PR #101, opened 2021-06-17
Initial config for OpenFn HAPI FHIR & OpenHIM

stuartc:
Here's a first draft of adding OpenFn Microservice into instant.

What we've done so far is port our [OpenFn/instant-demo](https://github.com/OpenFn/instant-demo) example into the instant repo.

I was hoping for some guidance and opinions about how we should integrate this into instant.

This goes around what to name the folder. It's currently called `openfn`.

@taylordowns2000 mentioned that adding the microservice image/app to `core` was a possibility.

It appears the structure that exists (other than core) is based around use case, like "here's a pre-baked and tested setup for a certain purpose" and core handles the rest.
And since the OpenFn Microservice can be used in lots of different ways (just like OpenHIM) it's feeling a bit weird having a folder named `openfn`.

So should we:

a) instead add microservice into `core`, and have another folder or change an existing one to add the specific configuration for working with HAPI FHIR via OpenHIM.
b) keep microservice in it's own folder but perhaps change the name to something other than `openfn`.
c) something else...

Thoughts?


taylordowns2000:
Hey @rcrichton , @MattyJ007 ! Keen on your thoughts here, re: our earlier conversations. Should a `microservice` instance with almost no config (think: a `project.yaml` without any jobs) go into `/core`?

In my mind, a lot depends on how we'd/you'd like folks to demo this. Would they clone the repo, add some config, and fire it up? Or would there be a hosted instance running somewhere (with core, including that generic `openfn-microservice` instance) that could then be configured to perform specific jobs via a request to a project configuration API (or some sort of UI for making this request.)

Eager to hear more.

mariusnaude:
Hi @taylordowns2000  and @stuartc . I've added this item to our internal backlog and we will get back to you on this topic.
Thank you.

taylordowns2000:
## Next steps:

- [ ] @stuartc and @taylordowns2000 to remove the "demo" configuration from this PR so that it's truly an infrastructure package that gets started with nothing but a hello world and/or "i'm alive" notification. it's `microservice` up and running as part of InstantOpenHIE, but it is awaiting configuration before it _does_ anything.
- [ ] @taylordowns2000 to submit a separate PR to the instant-openhie docs site which outlines how to configure a `microservice` instance via that `project.yaml` file _or_ via REST.
- [ ] someone (@MattyJ007 ?) to merge this PR
- [ ] once merged, @stuartc to update the https://github.com/openFn/instant-demo repo to align it with https://github.com/jembi/who-covid19-surveillance-package . see below for details

## instant-demo repo

1. should be renamed to indicate the use case it's demoing
2. should include configuration (via project.yaml or REST config) for a generic `microservice` instance that gets booted up when starting Instant OpenHIE.
3. should include fixtures and tests that demonstrate the particular functionality/use-case... in this case, how forms submitted via CommCare, Kobo, Ona, etc., are manipulated to adhere to FHIR specs and then sent to HAPI-FHIR via a channel.

PR #103, opened 2021-06-29
Add elastic stack infrastructure packages

rcrichton:
DIO-22


MattyJ007 (approved):
LGTM - not 100% sure what's happening with the extra pipeline mounting but all the services were accessible - I guess we'll get into the details of this package when we use it 👍🏼

PR #110, opened 2021-08-12
Dio 54 analytics package tests

bradsawadye:


elastic-analytics/features/elastic-analytic.js
```diff
@@ -0,0 +1,65 @@
...
+Then('the JS Report service should be up and running', async () => {
+  console.log('\n\nChecking the status of the JS Report service\n\n')
+
+  try {
+    await sendRequest(JS_REPORT_API_PORT)
+  
+    console.log('JS Report service is up\n')
+  } catch (error) {
+    if (error.response.status != 500) {
+      console.log('JS Report service is up\n')
```
MattyJ007, line 28: @bradsawadye Do we expect errors? Why not fail the test if there's an issue?

elastic-analytics/features/elastic-analytic.js
```diff
@@ -0,0 +1,65 @@
...
+Then('the ES analytics service should be up and running', async () => {
+  console.log('\n\nChecking the status of the ES Analytics Service\n\n')
+
+  try {
+    const result = await sendRequest(ES_ANALYTICS_API_PORT)
+
+    console.log('The ES Analytics service is up\n')
+  } catch (error) {
+    if (error.response.status != 500) {
+      console.log('The ES Analytics service is up\n')
```
MattyJ007, line 44: Same as above

elastic-analytics/features/elastic-analytic.js
```diff
@@ -0,0 +1,65 @@
...
+Then('the Kibana shpuld be up and running', async () => {
+  console.log('\n\nChecking the status of the Kibana Service\n\n')
+
+  try {
+    const result = await sendRequest(KIBANA_API_PORT)
+
+    console.log('The Kibana service is up\n')
+  } catch (error) {
+    if (error.response.status != 500) {
+      console.log('The Kibana service is up\n')
```
MattyJ007, line 60: Same as above

MattyJ007 (changes requested):
@bradsawadye The tests here look pretty good - but could you also please add some tests for the pipeline.
The analytics and pipeline packages won't be used independently AFAIK. Either way - they could both use up tests

elastic-analytics/features/elastic-analytic.js
MattyJ007: @bradsawadye

bradsawadye: @MattyJ007, I did it this way to cater for 401 errors. Some services require authentication.

MattyJ007: @bradsawadye Could you please add a comment in the code to explain this. Currently, it looks dodgy because you're hiding an error.

Also, can you add something to the console logs when it is a weird status code i.e. `console.log('JS Report service is up and responded with ${error.response.status}\n') `

MattyJ007 (approved):
LGTM

PR #122, opened 2021-09-30
Ohie 500 jvm tuning hapi fhir reduce memory usage

Ken473:
# If applied, this commit will:
Limit HAPI FHIR containers to use a maximum of 2GB

# Explain why this change is being made
To prevent cost overruns with containers

# Provide links to any relevant tickets, articles or other resources
Ticket number is OHIE-500.
links are: https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/ 
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
https://linuxhint.com/docker_compose_memory_limits/

### DIFF
--- a/core/docker/docker-compose.yml
+++ b/core/docker/docker-compose.yml
@@ -43,7 +43,7 @@ services:
         target: /instant
     depends_on:
       - mysql
-
+    mem_limit: 2g
   mysql:
     container_name: hapi-mysql
     image: mysql:5.7
--- a/core/kubernetes/main/hapi-fhir/hapi-fhir-server-deployment.yaml
+++ b/core/kubernetes/main/hapi-fhir/hapi-fhir-server-deployment.yaml
@@ -23,6 +23,7 @@ spec:
       containers:
         - name: hapi-fhir-server
           image: hapiproject/hapi:latest
+          mem_limit: 2g
           env:
             - name: spring.datasource.url
               value: jdbc:mysql://hapi-fhir-mysql-service:3306/hapi
--- a/core/kubernetes/main/hapi-fhir/hapi-fhir-server-deployment.yaml
+++ b/core/kubernetes/main/hapi-fhir/hapi-fhir-server-deployment.yaml
@@ -23,7 +23,9 @@ spec:
       containers:
         - name: hapi-fhir-server
           image: hapiproject/hapi:latest
-          mem_limit: 2g
+          resources:
+            limits:
+              memory: "2Gi"
           env:
             - name: spring.datasource.url
               value: jdbc:mysql://hapi-fhir-mysql-service:3306/hapi


MattyJ007 (changes requested):
@Ken473 
Your solution works, however, I think this limitation should be on HAPI-FHIR itself, not its container. The Instant OpenHIE is showing users how these systems can be connected - it would be more useful for real implementations (not necessarily using Docker) to see the JVM config mechanism.
Hence, in the ticket, I linked in how this was done in another part of the project: 
https://github.com/openhie/instant/blob/master/elastic-pipeline/docker/jvm.options#L6

Ken473:
> @Ken473 Your solution works, however, I think this limitation should be on HAPI-FHIR itself, not its container. The Instant OpenHIE is showing users how these systems can be connected - it would be more useful for real implementations (not necessarily using Docker) to see the JVM config mechanism. Hence, in the ticket, I linked in how this was done in another part of the project: https://github.com/openhie/instant/blob/master/elastic-pipeline/docker/jvm.options#L6

duly noted @MattyJ007, So you'd like the limitation to be applied to the JVM itself, not the container as we could have instant implementations without docker. Do you want to apply this to Kubernetes?

MattyJ007:
Kubernetes should be limited in the same way (on the JVM)

core/docker/docker-compose.yml
```diff
@@ -37,13 +37,15 @@ services:
       - hapi.fhir.allow_external_references=true
       - hapi.fhir.max_page_size=50
       - hapi.fhir.default_page_size=20
+      - "JAVA_TOOL_OPTIONS=-Xmx2g"
+      - "CATALINA_OPTS=-Xmx2g"
     volumes:
       - type: volume
         source: instant
         target: /instant
+
```
MattyJ007, line 46: Remove random newline - the depends_on is part of the hapi-fhir container

core/docker/docker-compose.yml
```diff
@@ -37,13 +37,15 @@ services:
...
+      - "JAVA_TOOL_OPTIONS=-Xmx2g"
+      - "CATALINA_OPTS=-Xmx2g"
     volumes:
       - type: volume
         source: instant
         target: /instant
+      
     depends_on:
       - mysql
-
```
MattyJ007, line 46: Add back the newline separator

core/docker/docker-compose.yml
```diff
@@ -62,4 +64,4 @@ volumes:
   hapi-mysql-config:
     external: true
   instant:
-    external: true
+    external: true
```
MattyJ007, line 67: Add newline to end of file - some interpreters expect it and github complains about it

core/kubernetes/main/hapi-fhir/hapi-fhir-server-deployment.yaml
```diff
@@ -23,6 +23,9 @@ spec:
       containers:
         - name: hapi-fhir-server
           image: hapiproject/hapi:latest
+          resources:
+            limits:
+              memory: "2Gi"
```
MattyJ007, line 28: Please implement the memory limit here the same way as in Docker (Env Vars)
We want the installations to be a similar as possible

MattyJ007 (changes requested):
How has the testing been going?
Can HAPI FHIR run with just 2gb memory with requests coming through?

core/docker/docker-compose.yml
```diff
@@ -37,13 +37,15 @@ services:
...
+      - "JAVA_TOOL_OPTIONS=-Xmx2g"
+      - "CATALINA_OPTS=-Xmx2g"
     volumes:
       - type: volume
         source: instant
         target: /instant
     depends_on:
       - mysql
-
+
```
MattyJ007, line 48: Please remove trailing whitespaces - it's bad practice to commit code clutter like this when a linter would have removed them

```suggestion

```

core/docker/docker-compose.yml
```diff
@@ -63,3 +65,4 @@ volumes:
     external: true
   instant:
     external: true
+
```
MattyJ007, line 68:
```suggestion

```

Ken473:
Hi @MattyJ007 thanks for the tip, will give you my testing results on slack. I have cleanup all the remaining spaces

PR #135, opened 2021-11-05
Support --only flag to only operate on specified packages

rcrichton:
This flag ignores resolving deps and acts on the specified packages only, in thge order they are provided.


MattyJ007 (approved):
Awesome :D

PR #136, opened 2021-11-10
Core package: Don't expose ports in prod mode and alway restart

rcrichton:
Adds support for a PROD=true env var which doesn't expose ports for use with a reverse proxy

### DIFF
--- a/core/docker/compose.sh
+++ b/core/docker/compose.sh
@@ -2,8 +2,16 @@
 
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
+if [ "$PROD" == "true" ]; then
+    devComposeParam=""
+else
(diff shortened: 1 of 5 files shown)


core/docker/compose.sh
```diff
@@ -2,8 +2,16 @@
 
 composeFilePath=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
+if [ "$PROD" == "true" ]; then
+    devComposeParam=""
```
MattyJ007, line 6: Perhaps add a printf to say it's in PROD mode?
```suggestion
    printf "\nRunning core package in PROD mode\n"
    devComposeParam=""
```

MattyJ007 (approved):
LGTM!
Nice work with the mongos. That separate .dev.yml fixed the restart!

PR #142, opened 2021-11-23
Switch hapi fhir from mysql to postgres

rcrichton:
Postgres is faster and recommended for production deployments. It's best to use this from the start.


MattyJ007 (approved):
Awesome!!

PR #146, opened 2021-11-24
Mount externally hosted custom packages

bradsawadye:
This adds functionality to download externally hosted packages and moutn them to the instantHIE instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,11 @@ For example, if you had downloaded the [who-covid19-surveillance-package](https:
 yarn docker:instant init core covid19surveillance -c="../who-covid19-surveillance-package"
 ```
 
-> We hope to support package urls soon
+Urls are supported. The custom package will be downloaded and then mounted to the instanst instance.
+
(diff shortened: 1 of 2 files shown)


MattyJ007 (commented):
@bradsawadye weren't we going to apply these changes in the `instant.ts` file?

bradsawadye:
@MattyJ007 yeah, I wanted to put the logic in the instant.ts but i realised that it makes sense to put it in the deploy.sh as this is where the mounting of locally hosted custom packages happens. The deploy script will then run the instant.ts file once the packages have been added to the instant volume.

MattyJ007:
@bradsawadye is this branch needed?

bradsawadye:
@MattyJ007 given that we shall not be using the deploy.sh, we no longer need the functionality in this pr

MattyJ007:
I'll merge it in for now since the functionality is there (in case we ever come back to deploy.sh) - we won't worry about it cause we'll soon remove the script

PR #148, opened 2021-11-30
Dio 86 custom packages mounting

bradsawadye:


goinstant/docker.go
```diff
@@ -187,3 +194,182 @@ func RunDockerCommand(commandSlice ...string) {
...
+		}
+		if _, err := io.Copy(dest, tarReader); err != nil {
+			fmt.Fprintln(os.Stderr, "Error in extracting tar file:", err)
+		}
+	}
+	pathToPackage = filepath.Join(".", packageName)
+	return
+}
+
+func retrieveGitRepo(gitUrl string) (pathToPackage string) {
```
michaelloosen, line 341: TODO: I will refactor this to use the same logic as the docker command, no need to duplicate this logic.

michaelloosen: Done

goinstant/docker.go
```diff
@@ -186,4 +193,166 @@ func RunDockerCommand(commandSlice ...string) {
...
+	httpRegex := regexp.MustCompile("http")
+	zipRegex := regexp.MustCompile(`\.zip`)
+	tarRegex := regexp.MustCompile(`\.tar`)
+
+	if gitRegex.MatchString(pathToPackage) {
+		pathToPackage = runCommand("git", []string{"clone", pathToPackage}...)
+	} else if httpRegex.MatchString(pathToPackage) {
+		resp, err := http.Get(pathToPackage)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Error in dowloading custom package", err)
```
michaelloosen, line 224: Should change these errors to Panic to prevent the method from returning an empty path and causing unexpected behaviour to the cli user

PR #153, opened 2021-12-08
Dio 108   custom package docs

mariusnaude:
Basic page added to link to the Instant use case packages.


MattyJ007 (changes requested):
@mariusnaude Could you possibly add a tiny summary of what each custom package does? Just a 2 sentences describing the package purpose so people don't need to go to a different page if it's not relevant to them

PR #160, opened 2021-12-13
DIO-109 go cli interactive functions

MattyJ007:


goinstant/prompts.go
```diff
@@ -81,57 +180,382 @@ func selectPackageDocker() {
...
+		fmt.Println("Stopping and Cleaning Up Electronic Medical Record...")
+		RunDirectDockerCommand([]string{"docker", "emr", "destroy"})
+		selectDefaultInstall()
+
+	case "Stop and Cleanup Health Management Information System":
+		fmt.Println("Stopping and Cleaning Up Health Management Information System...")
+		RunDirectDockerCommand([]string{"docker", "hmis", "destroy"})
+		selectDefaultInstall()
+
+	case "Stop and Cleanup Health Worker":
```
michaelloosen, line 539: Is this not duplicating the logic under "Stop and Cleanup Workforce"?

michaelloosen (changes requested):

goinstant/prompts.go
```diff
@@ -81,57 +180,382 @@ func selectPackageDocker() {
...
+		selectCustomOptions()
+	}
+}
+
+func executeCommand() {
+	startupCommands := []string{"docker", customOptions.startupAction}
+	if customOptions.startupPackages != nil && len(customOptions.startupPackages) > 0 {
+		startupCommands = append(startupCommands, customOptions.startupPackages...)
+	} else {
+		fmt.Println("No startup package specified, cannot start.")
```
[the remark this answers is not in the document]
MattyJ007: Perhaps remove it for consistency. I don't think anyone will use the init all packages option in reality - but the destroy all is very useful (especially in testing) sometimes I start up components in odd projects then just want to clean up everything without much typing 🙃

goinstant/prompts.go
MattyJ007: Health Workforce and Health Worker are different components

michaelloosen: Indeed, but both have exactly the same docker command, so what am I missing here? 😁

michaelloosen: Indeed, so they should have different docker commands right? 😁

MattyJ007: lol, I didn't expand the whole file :p thought you were confused

michaelloosen: Cool, thanks for correcting it 🙂

PR #169, opened 2022-01-14
Allow go cli to be configurable

rcrichton:
The go cli is now configurable via config.yml to allow different distributions of the cli to be created with a different set of packages and default config options.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:fermium-buster
+FROM node:fermium-buster AS base
 
 WORKDIR /instant
 
@@ -30,6 +30,11 @@ ADD package.json .
 ADD yarn.lock .
 RUN yarn
 
-ADD . .
+# add entrypoint script
+ADD instant.ts .
 
 ENTRYPOINT [ "yarn", "instant" ]
+
+FROM base as instant-build
+# Add default instant OpenHIE packages
+ADD . .
--- a/goinstant/goinstant.go
+++ b/goinstant/goinstant.go
@@ -48,7 +48,10 @@ var customOptions = customOption{
 func stopContainer() {
 	commandSlice := []string{"stop", "instant-openhie"}
 	suppressErrors := []string{"Error response from daemon: No such container: instant-openhie"}
-	runCommand("docker", suppressErrors, commandSlice...)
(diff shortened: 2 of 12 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:fermium-buster
+FROM node:fermium-buster AS base
 
 WORKDIR /instant
```
MattyJ007, line 3: Perhaps we could name this something more generic to allow more logical reuse?
`/packages` ?
`/components` ?

goinstant/goinstant.go
```diff
@@ -25,7 +25,7 @@ type Config struct {
 	Image              string    `yaml:"image"`
 	DefaultEnvironment string    `yaml:"defaultEnvironment"`
 	Packages           []Package `yaml:"packages"`
-	DisableKuberneses  bool      `yaml:"disableKubernetes"`
+	DisableKubernetes  bool      `yaml:"disableKubernetes"`
```
MattyJ007, line 28: Good catch 😉

Dockerfile
rcrichton: Agreed, but that might break packages that expect config to be in the instant volume. We'd have to re-look at how the volume is mounted before we make changes. I'm actually thinking that we need to do some thinking and re-work of the how the instant volume works. I think we need to ensure the latest from the image is always loaded in there. I think we should look at this as a separate piece of work.

goinstant/goinstant.go
```diff
@@ -30,7 +48,10 @@ var customOptions = customOption{
 func stopContainer() {
 	commandSlice := []string{"stop", "instant-openhie"}
 	suppressErrors := []string{"Error response from daemon: No such container: instant-openhie"}
-	runCommand("docker", suppressErrors, commandSlice...)
+	_, err := runCommand("docker", suppressErrors, commandSlice...)
+	if err != nil {
+		gracefulPanic(err, "")
```
Mark-Labuschagne, line 53: This call to gracefulPanic results in an infinite loop on error. I fixed this in a separate PR.

Mark-Labuschagne (approved):

MattyJ007 (approved):

PR #173, opened 2022-01-17
Fix error handling bugs in the go CLI.

Mark-Labuschagne:


MattyJ007 (approved):
Possibly a few bugs here relating to the instant volume but I think we'll address those with unit tests

PR #174, opened 2022-01-19
Plat 37 go cli unit tests

Mark-Labuschagne:


goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+	for _, tt := range testCases {
+		runCommand = tt.runCommandMock
+		tt.mockServer()
+		unzipPackage = tt.unzipPackageMock
+		untarPackage = tt.untarPackageMock
+
+		t.Run(tt.testInfo, func(t *testing.T) {
+			err := mountCustomPackage(tt.pathToPackage)
+			if err == nil && tt.wantErr {
+				t.Fatal("Expected error - '" + tt.errorString + "' but got nil")
```
Mark-Labuschagne, line 576: If we're calling t.Fatal(), we're exiting the running of all test cases. t.Errorf(), on the other hand, will return an error for just that test case, but still run the other test cases. So, if we fail one test case, do we not want to run the other test cases?

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+		unzipPackageMock func(zipContent io.ReadCloser) (pathToPackage string, err error)
+		untarPackageMock func(tarContent io.ReadCloser) (pathToPackage string, err error)
+	}{
+		{
+			pathToPackage:  "http://test:8080/test",
+			mockServer:     func() {},
+			errorString:    "Error in downloading custom package",
+			wantErr:        true,
+			testInfo:       "mountPackage - should return error when downloading custom package fails",
+			runCommandMock: func(commandName string, suppressErrors []string, commandSlice ...string) (pathToPackage string, err error) {
```
Mark-Labuschagne, line 401: I think naming this function 'runCommandMock()' is misleading. Maybe 'mockRunCommand()' would be better?

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+		targetLauncher       string
+		packages             []string
+		customPackagePaths   []string
+		instantVersion       string
+	}
+
+	testCases := []struct {
+		startupCommands []string
+		expectedResults resultStruct
+		testInfo        string
```
Mark-Labuschagne, line 138: Some of my tests use the default field 'name' as what you have as 'testInfo'. We should decide on a convention and stick to it.

Mark-Labuschagne: Same with default field 'args' for specifying arguments to the function being tested. Let's decide on a convention and stick to it.

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+			testInfo: "Extract commands test 4 - should return the expected commands",
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.testInfo, func(t *testing.T) {
+			environmentVariables, deployCommand, otherFlags, packages, customPackagePaths, instantVersion, targetLauncher := extractCommands(tt.startupCommands)
+
+			if !assert.Equal(t, environmentVariables, tt.expectedResults.environmentVariables) {
+				t.Fatal("ExtractCommands should return the correct environment variables")
```
Mark-Labuschagne, line 199: See later comment about t.Fatal(). Not going to mention that at every instance, but think about whether or not we want all test cases to quit if we fail one test case.

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+			}
+			if !assert.Equal(t, deployCommand, tt.expectedResults.deployCommand) {
+				t.Fatal("ExtractCommands should return the correct deploy command")
```
Mark-Labuschagne, line 202: It would be nice to have some context on these errors.

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			OsCreate = tt.osCreate
+			IoCopy = tt.ioCopy
+
+			if err := createZipFile(tt.args.file, tt.args.content); (err != nil) != tt.wantErr {
+				t.Errorf("createZipFile() error = %v, wantErr %v", err, tt.wantErr)
+				log.Println(tt.name, "failed!")
```
Mark-Labuschagne, line 302: I'll change my instances of log.Println() to use t.Log instead.

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+			}
+
+			t.Log(tt.testInfo + " passed!")
+		})
+	}
+}
+
+func Test_mountPackage(t *testing.T) {
+	defer gock.Off()
+	defer resetMountPackageMocks(runCommand, unzipPackage, untarPackage)
```
Mark-Labuschagne, line 383: Not quite sure what you're trying to do here?

If it's resetting the variables for use outside of testing, shouldn't we be doing that everywhere? <-- this obviously depends on the plan for running these tests.

goinstant/docker_test.go
```diff
@@ -0,0 +1,948 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			RunCommand = tt.mockRunCommand
+			MountCustomPackage = tt.mockMountCustomPackage
+
+			if err := RunDeployCommand(tt.args.startupCommands); (err != nil) != tt.wantErr {
+				t.Errorf("RunDeployCommand() error = %v, wantErr %v", err, tt.wantErr)
```
Mark-Labuschagne, line 944: do t.Log("test passed")

Mark-Labuschagne: This is for myself...

goinstant/docker_test.go
bradsawadye: The other tests for functions that also use the functionality that is mocked were failing. The mocked functions have to be reset

goinstant/docker_test.go
bradsawadye: @Mark-Labuschagne. Thanks for the feedback man... Calling `t.Fatal` does not stop the other test cases from executing, it only stops the execution of that particular test case. `t.Errorf` marks the test case as failed, but does not stop the execution of that test case. Logic that should not be executed when the test fails (like the success messages we log out) will be executed when we use `t.Errorf`.

goinstant/docker_test.go
bradsawadye: I see you changed from testInfo to name. Will update for consistency

goinstant/docker_test.go
bradsawadye: If there is a failure `assert.Equal` will log out the values being compared

goinstant/docker_test.go
Mark-Labuschagne: Ahh cool, sorry for my confusion. What I think we should try then is do an else check that prints the test case passed. For instance
`
if (err != nil) != tt.wantErr {
				t.Errorf("unzipPackage() error = %v, wantErr %v", err, tt.wantErr)
				t.Log(tt.name, "failed!")
			} else {
				t.Log(tt.name, "passed!")
			}
`
This would print out which test case failed as well, making it a bit easier to spot the fail case...

Mark-Labuschagne: I'm checking now that this actually only works if we use log.Println() for printing... t.Log only prints information from the fail case

michaelloosen (approved):

PR #175, opened 2022-01-20
Fix instant.ts and go cli to cater for swarm mode

michaelloosen:
This will allow instant to run against a docker swarm
PLAT-57


MattyJ007 (approved):
LGTM

PR #176, opened 2022-01-21
Fix logic for targeting the correct environment

michaelloosen:
The previous logic would not allow for swarm mode to be used in non-interactive mode.
PLAT-67

### DIFF
--- a/goinstant/cli.go
+++ b/goinstant/cli.go
@@ -4,8 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-
-	"github.com/fatih/color"
 )
 
 func CLI() error {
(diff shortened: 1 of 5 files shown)


bradsawadye (approved):

goinstant/cli.go
```diff
@@ -74,7 +61,15 @@ Commands:
 			err = loadIGpackage(startupCommands[1], startupCommands[2], params)
 		}
 	default:
-		fmt.Println("The deploy command is not recognized: ", startupCommands)
+		if len(startupCommands) < 2 {
+			fmt.Println("The deploy command is not recognized: ", startupCommands)
```
Mark-Labuschagne, line 65: fmt.Println() adds a space by itself. So, the space at 'recognized: ' is not necessary.

PR #179, opened 2022-01-25
Plat 37 go cli integration testing

Mark-Labuschagne:


goinstant/features/service_destroy.feature
```diff
@@ -0,0 +1,4 @@
+Feature: destroy service
+  Scenario: destroy the service                   
+    When the service is destroyed
+    Then check the service is destroyed
```
bradsawadye, line 4:
```suggestion
    Then check the service is destroyed

```

bradsawadye: @Mark-Labuschagne Please go through this - https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

goinstant/features/service_down.feature
```diff
@@ -0,0 +1,4 @@
+Feature: bring service down
+  Scenario: bring the service down
+    When the service is brought down
+    Then check the service is brought down
```
bradsawadye, line 4:
```suggestion
    Then check the service is brought down

```

goinstant/features/service_init.feature
```diff
@@ -0,0 +1,4 @@
+Feature: initialise service
+  Scenario: initialise the service
+    When the service is initialised
+    Then check the service is initialised
```
bradsawadye, line 4:
```suggestion
    Then check the service is initialised

```

goinstant/features/service_options.feature
```diff
@@ -0,0 +1,4 @@
+Feature: configure service
+  Scenario: configure the service
+    When the service config options are passed
+    Then check the service config options are passed
```
bradsawadye, line 4:
```suggestion
    Then check the service config options are passed

```

goinstant/features/service_up.feature
```diff
@@ -0,0 +1,4 @@
+Feature: bring service up
+  Scenario: bring the service up
+    When the service is brought up
+    Then check the service is brought up
```
bradsawadye, line 4:
```suggestion
    Then check the service is brought up

```

goinstant/README.md
```diff
@@ -48,3 +48,13 @@ bash ./buildreleases.sh
 To build releases, create an instant tag and a release, the GitHub actions will build the code after creation of the release and add the binaries to the assets of the release if the GO CLI build succeeds.
 
 Check [here](https://github.com/openhie/instant/actions/new) to review the output of the build and status of the binary deploy.
+
+### Testing
+
+The [godog library](https://github.com/cucumber/godog), which provides us with the [Cucumber Framework](https://cucumber.io/) is used for functional tests of the Go CLI. To run the tests install godog and then navigate to the `goinstant` root folder and run the command below.
```
Mark-Labuschagne, line 54: The one sentence here is a little difficult to read... try replacing with the following:

"To run the tests, install godog, then navigate to the `goinstant` root folder. Finally, run the command below."

goinstant/features/service_commands.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Test Deploy Commands
+  Scenario: Initialise Core Service
+    When the command "init core -t=docker" is run
```
michaelloosen, line 3: Specifying the target as docker should no longer be required here ([bugfix](https://github.com/openhie/instant/pull/180/files) has been merged), so I think we should have some tests using it and some not, to ensure we cover that code as well?

goinstant/godogs_test.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+var (
+	binaryFilePath string
+	logs           string
+	directoryNames []string
+)
+
+func theCommandIsRun(command string) error {
+	res, err := runTestCommand(binaryFilePath, strings.Split(command, " ")...)
```
michaelloosen, line 25: Hmm, this seems prone to failure, I know it's highly unlikely, but a subpart from a command should be able to have a space in it, so is it possible to change the command parameter to be a slice of type string instead of needing to do the `strings.Split` here?

goinstant/godogs_test.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+		}
+	}
+
+	// Cleanup downloaded packages post test
+	deleteContentAtFilePath([]string{".", "features"}, directoryNames)
+	return nil
+}
+
+func compareStrings(inputLogs, expectedOutput string) error {
```
michaelloosen, line 65: Can we please give this function a better name?

goinstant/godogs_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	os.Exit(0)
+}
+
+func buildBinary() string {
+	_, err := runTestCommand("/bin/sh", filepath.Join(".", "features", "build-cli.sh"))
+	if err != nil {
+		panic(err)
+	}
+	_, err = os.Stat(filepath.Join(".", "features", "test-platform-linux"))
```
michaelloosen, line 99: Why are we checking a specific OS binary file here? Is this just to check if the `build-cli.sh` script succeeded?

michaelloosen (changes requested):

goinstant/features/service_commands.feature
Mark-Labuschagne: This would only work if we were calling the commands through the go functions. But, what actually happens is that we take that command and pass it through to a cloned version of the function runCommand (named runTestCommand in godogs_test.go), thereby bypassing extractCommands (where the default target launcher is checked)

goinstant/godogs_test.go
Mark-Labuschagne: On the second name already, do you have a suggestion?

goinstant/godogs_test.go
Mark-Labuschagne: I'll speak under correction, but I don't think it is possible. I think it's just how gherkin passes the info through...

goinstant/godogs_test.go
bradsawadye: Yeah, its for checking whether `build.cli` has succeeded

goinstant/godogs_test.go
michaelloosen: Okay, guess we'll have to live with it and figure out another solution if it ever becomes a problem.

michaelloosen (approved):

PR #180, opened 2022-01-26
Fix default target launcher

michaelloosen:
The default target launcher wasn't being set when running in non-interactive mode, this will set it to the default that's in the config file.
PLAT-75


MattyJ007 (approved):
LGTM

PR #188, opened 2022-02-14
Plat 69 instant volume update fix

Mark-Labuschagne:
@rcrichton , we should evaluate if we want the instant volume to hang around in the first place. As it stands, this PR will update package data when doing an up command.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,7 +10,19 @@ goinstant
 
 # Docs
 docs
+kubernetes.md
+README.md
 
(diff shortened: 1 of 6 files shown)


rcrichton:
Hey @Mark-Labuschagne cool this is looking good. What do you mean about if we want the volume to hang around? The volume needs to exist after the orchestration container exits as some container that packages start rely on the instant volume. I don't think I understand what you are asking.

PR #189, opened 2022-02-16
Add some base error handling so we don't swallow top level thrown errors

rcrichton:
This change will cause the Go CLI to raise an error if the instant.ts script encounters an error. We should discuss if this is appropriate or if the CLI should rather handle the error and print out something to the user.

### DIFF
--- a/instant.ts
+++ b/instant.ts
@@ -132,7 +132,7 @@ const logPackageDetails = (packageInfo: PackageInfo) => {
 }
 
 // Main script execution
-;(async () => {
+const main = async () => {
   const allPackages = getInstantOHIEPackages()
   console.log(
     `Found ${Object.keys(allPackages).length} packages: ${Object.values(
@@ -287,4 +287,13 @@ const logPackageDetails = (packageInfo: PackageInfo) => {
       await runTests(features)
     }
   }
-})()
+}
+
+// Entry point IIFE with base error handling
+;(async () => {
+  try {
+  await main()
+} catch (error) {
+  console.log(error)
+  process.exit(1)
+}})()


rcrichton:
I though it might be better for the Go CLI to not exit when the orchestration container fails so i added a commit to do that. Now it will take the user back to a prompt and print out the error.

goinstant/docker.go
```diff
@@ -189,11 +190,13 @@ func RunDeployCommand(startupCommands []string) error {
 		}
 	}
 
-	fmt.Println("\nRun Instant OpenHIE Installer Container")
+	fmt.Println("\nRunning orchestration container")
 	commandSlice = []string{"start", "-a", "instant-openhie"}
 	_, err = RunCommand("docker", nil, commandSlice...)
 	if err != nil {
-		return err
+		color.Red("\nError: Failed while running orchestration container. Look for error output above.\n")
```
Mark-Labuschagne, line 197: Good thinking, but, since we don't know at which point the error is happening at RunCommand(), I think we should log the error coming from RunCommand().

```suggestion
		color.Red("\nError: Failed while running orchestration container, received err: " + err.Error() + "\n")
```

Mark-Labuschagne (changes requested):

goinstant/docker.go
rcrichton: For a user I think it's more important to point them to the output of the container as that will help them figure out what's wrong, but I also get we should print more detail about the error object itself for dev/debug purposes. I've ended up with something like this. What do you think?

```
Running orchestration container
         > yarn run v1.22.5
         > $ ts-node instant.ts init -t swarm
         > Found 9 packages: client, elastic-analytics, elastic-pipeline, fhir-datastore-hapi-fhir, interoperability-layer-openhim, message-bus-helper-hapi-proxy, message-bus-kafka, reverse-proxy-nginx-config-importer, reverse-proxy-nginx
         > Target environment is: swarm
         > Error: Package core does not exist or the metadata is invalid
         >     at resolveDeps (/instant/instant.ts:103:13)
         >     at /instant/instant.ts:110:19
         >     at Array.forEach (<anonymous>)
         >     at resolveDeps (/instant/instant.ts:109:43)
         >     at /instant/instant.ts:119:47
         >     at Array.forEach (<anonymous>)
         >     at orderPackageIds (/instant/instant.ts:118:20)
         >     at /instant/instant.ts:199:26
         >     at step (/instant/instant.ts:33:23)
         >     at Object.next (/instant/instant.ts:14:53)
         > info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Error: Failed while running orchestration container, check output from Orchestration container above. (Underlying error: Error waiting for Cmd. error Command failed with exit code 1.
: exit status 1)
```

Mark-Labuschagne: Seems good to me :+1:

MattyJ007 (approved):
LGTM

PR #190, opened 2022-02-18
Move the removing of the instant volume to after package deploy.

Mark-Labuschagne:
To ensure that the insant volume does not hang around.
PLAT-69


MattyJ007 (approved):
👍🏼

PR #193, opened 2022-02-23
Rename build releases github action and add github action for running

Mark-Labuschagne:
tests.
To run tests, ensuring that tests pass on each git push.
PLAT-124


goinstant/docker_test.go
```diff
@@ -857,7 +857,7 @@ func TestRunDeployCommand(t *testing.T) {
 			args: args{
```
michaelloosen, line 857: The name(description) should also reflect this change.

.github/workflows/GoCliTests.yaml
```diff
@@ -0,0 +1,27 @@
+name: Run Go CLI unit tests and Godog functional tests
+
+on:
+  push:
```
michaelloosen, line 4: Isn't there a better way to do this to monitor only changes to the goinstant directory? Or is this why we need a git submodule for the GO CLI?

michaelloosen: As per discussions, this can be done in a future task.

michaelloosen (approved):

PR #197, opened 2022-03-18
Plat-186 Change docker base image

tumbledwyer:


Dockerfile
```diff
@@ -1,40 +1,15 @@
...
-
-ENTRYPOINT [ "yarn", "instant" ]
-
-FROM base as instant-build
-# Add default instant OpenHIE packages
-ADD . .
+# install kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+RUN chmod +x ./kubectl
+RUN mv ./kubectl /usr/local/bin/kubectl
```
MattyJ007, line 15: @tumbledwyer no newline at end of file :( 
Toggle your VS code setting

MattyJ007 (changes requested):
@tumbledwyer There are changes missing:
- package.json
  the launch specific stuff needs to be removed - leaving behind the current test framework content - this will apply to yarn.lock automatically
-  add a `instant-test-harness` dir
  move package.json, yarn.lock, test.sh, and PackageTest.dockerfile into dir
- PackageTest.dockerfile
  update the test base image to gallium to match package-starter-kit
- .dockerignore
  we need to make sure we don't accidentally copy over a package.json into the docker image.
  this is a preventative measure as package.json is now in the test dir
- Remove instant.ts
  it now lives in package-starter-kit.
- Remove deploy.sh
  not used. go cli took all it's responsibilities
- Remove github workflow
  these all require the goinstant dir
- .prettierrc.json
  this file was supposed to be the linter for instant.ts (it should apply to all the test files in features dirs as well) this needs an update whatever the case.
- Remove travis.yml
  Boyscout - OpenHIE doesn't have a paid travis so we can't use it this :( :(

MattyJ007 (changes requested):
I missed the `run-cli.sh` file - this also needs an update

Dockerfile
tumbledwyer: done

MattyJ007 (approved):
LGTM

PR #210, opened 2022-11-23
Always get the latest Go CLI version, for your OS

kaapstorm:
This changes the `get-cli.sh` script always to get the version of the Go CLI tool tagged as "latest" (which is currently version 0.9.0).

It also allows the user to state which OS they want it for, instead of downloading all of them.


kaapstorm:
(I'm not sure who to ping for a review. @rcrichton ?)

